PR #7, opened 2016-01-22
typo

slajax:
It's actually only a 1 character typo fix.

Not sure why github thinks line 26 changed. 

Perhaps we need to add new lines to the end of these files?


marshallswain:
Thanks!

PR #16, opened 2016-01-28
Documentation for middleware, routing and versioning

daffl:
Getting closer!


daffl:
This should explain many of the common questions like how to do nested routes or using `express.Router` and other pitfalls between routes and service names.

PR #29, opened 2016-02-19
Update mongoose.md

kulakowka:
Example of validation did not work. Probably because the libraries are called similar (`validator` and `validator.js`).

I fixed an example. I just tested it on a new application. Now email validation is working correctly.


ekryski:
Nice. Ok glad that works no! Thanks @kulakowka!

PR #36, opened 2016-02-23
Change all ES6 module imports back to CommonJS to work with Node5 out…

daffl:
… of the box


ekryski:
:shipit: Looks good. I had migrated a bunch of the other examples over as well.

PR #38, opened 2016-02-23
correct basic example with missing collection statement

marco4net:


ekryski:
Woot! Thanks @marco4net!

PR #43, opened 2016-03-04
Initial version of the angular2 guide

harangue:
I created an Angular 2 guide. There's not a ton of feathers-specific content in there because it's so simple to setup. Instead, I detail the service structure that I recommend for feathers and a few of the strange Angular-y issues users are likely to run in to.

Really my code is nothing that anyone past an Angular novice couldn't figure out for themselves, so I kept it pretty lightweight and tried to explain things thoroughly.

I didn't edit super heavily so there could be some errors in it as well. Let me know what you think.


ekryski:
@harangue Man this is great Jack! Thank you very much for taking the time to write this. If you're up for it I'd love to get you to do a guest submission on medium for this in the future. Should be able to just copy paste the content more or less but I want you to have the credit.

I have a couple minor edits I'll make after we merge in but this is great!

PR #44, opened 2016-03-08
using NeDB instead of MongoDB

diego-vieira:


daffl:
That is indeed true. Thank you!

PR #47, opened 2016-03-08
Unnecessary install of pg driver

diego-vieira:
Since the user will choose between of the the available drivers from sequelize, installing pg pg-hstore is useless.

### DIFF
--- a/databases/sequelize.md
+++ b/databases/sequelize.md
@@ -14,7 +14,7 @@ npm install --save sqlite3
 npm install --save tedious // MSSQL
 ```
 
-> **ProTip:** Full list of available drivers available at [Sequelize documentation](http://docs.sequelizejs.com/en/latest/docs/getting-started/). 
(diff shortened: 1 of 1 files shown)


ekryski:
Hmm I think this is more of an issue with the docs being confusing. Depending on the database you are using you do need to install the db driver as well. Maybe we should just have examples for each database with sequelize?

diego-vieira:
I thought to the same, but since there is tip telling you to install the DB driver, I didn't thought about adding it.

Perhaps like sequelize docs
```
$ npm install --save sequelize

# And one of the following:
$ npm install --save pg pg-hstore
$ npm install --save mysql // For both mysql and mariadb dialects
$ npm install --save sqlite3
$ npm install --save tedious // MSSQL
```

diego-vieira:
@ekryski should be good now

ekryski:
:boom: Awesome! That's perfect.

PR #48, opened 2016-03-08
Tip for getting json response instead of HTML

diego-vieira:
While prototyping, first thing most devs do is fire postman or curl and start hitting the rest api and might not know that they need to set `Accept: application/json` for feathers to send a json response.

https://github.com/feathersjs/feathers-mongoose/issues/56
https://feathersjs.slack.com/archives/general/p1457449015000948

### DIFF
--- a/middleware/error-handling.md
+++ b/middleware/error-handling.md
@@ -22,6 +22,9 @@ app.use(error({
     }
 }))
 ```
+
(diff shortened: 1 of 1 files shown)


ekryski:
Totally right. Thanks I had been meaning to add this!

PR #50, opened 2016-03-09
Added new guide for creating a personalized user endpoint

harangue:
The name and location in the guide might need revision, but this page details my method of dealing with a user querying their private information.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -62,6 +62,7 @@
    * [Migrating to Feathers 2](guides/migrating.md)
    * [Server Side Rendering](guides/server-side-rendering.md)
    * [Creating a plugin](guides/creating-a-plugin.md)
+   * [Creating a /me endpoint](guides/me-endpoint.md)
    * [React](guides/react.md)
    * [React Native](guides/react-native.md)
    * [jQuery](guides/jquery.md)
--- a/guides/me-endpoint.md
+++ b/guides/me-endpoint.md
@@ -1,0 +1,76 @@
+# Creating a /me endpoint
+
+If you're using [feathers-authentication](authentication/readme.md), it's likely that you'll have private information on your user resources that should be available only to the people it belongs to and no one else.
+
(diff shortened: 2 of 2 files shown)


ekryski:
Sorry I haven't had time to look at this. When I did a quick skim I think we can simplify it so that you don't even need to create a separate endpoint to manage. Instead I would probably call this guide something like "Adding additional fields to the current user". We _should_ be able to use a populate hook to just add data conditionally if the resource requested is the current user, otherwise just bypass the hook.

harangue:
I dunno, I kind of like having the separation. It's how Facebook Graph Search does it, with a `/me` endpoint. But I understand that it doesn't really buy into the whole hooks scheme. I might just throw this  up on my personal blog if you'd prefer that.

ekryski:
@harangue Absolutely put it up on your blog regardless! At the very least we will link to it. Let me post up something today/tomorrow for what I was thinking and then we can discuss. Would also like others to weigh in on it. Just because I don't do it that way doesn't mean other won't prefer that method.

ekryski:
Coming back to this.... I think because you get the user's ID in the response from successful auth and also in the JWT payload (which you can decode it on the client) there really isn't a need for an explicitly `/users/me` endpoint. You already have the ID to do `/users/:id` and you can easily just use a hook to say only return X data unless I am the owner of this resource.

This really is more related to sanitizing data with hooks (#134). So I think with a few tweaks this actually is a great use case for showcasing how hooks can be used. I would just prefer to not have a special endpoint because you don't really need it. It's kind of an anti-pattern in Feathers.

@harangue does that make sense? Do you want to change it up? Otherwise I will close this PR and re-purpose some of your content from this PR for a new one against #134.

harangue:
@ekryski Yep, I think hooks is the most Feather-sy solution as well. I wanted an API in the style of some other ones I've seen so this pattern worked pretty well for me, but I'm fine with it not being pulled.

PR #54, opened 2016-03-10
fixing typo

jamesjnadeau:


jamesjnadeau:
ps I like this workflow!

ekryski:
Awesome thanks @jamesjnadeau! Glad this makes it easy. I think it's pretty ideal to allow people to contribute to the doc.

PR #56, opened 2016-03-15
Update react.md

kulakowka:
Fixed syntax highlighting for JSX code snippets.


ekryski:
I think that should work in gitbook! Thanks @kulakowka

PR #79, opened 2016-03-18
fixed word error: compromised --> comprised

tlossen:


ekryski:
Ha ah. A little different meaning. Thanks @tlossen!

PR #80, opened 2016-03-18
Update primus.md

pierrenel:
fixed a typo - I'm pretty sure you meant const primus, not const primuse.


ekryski:
Thanks @pierrenel!

PR #84, opened 2016-03-19
Fixed 404

katopz:
mongoosejs.com -> http://mongoosejs.com/


corymsmith:
Nice catch, thanks @katopz!

PR #95, opened 2016-03-22
Update Parse

swarthout:
Parse now includes Realtime from Server and has React Native Support
links:
Live Queries: http://blog.parse.com/announcements/parse-server-goes-realtime-with-live-queries/

React Native: https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/README.md 
(read "Using Parse on Different Platforms")

### DIFF
--- a/why/vs/readme.md
+++ b/why/vs/readme.md
@@ -166,7 +166,7 @@ If you find something invalid or out of date in the comparisons, please [create 
             <td>❌ 🤓</td>
             <td>❌ 🤓</td>
             <td>✅</td>
-            <td>❌ 🤓</td>
+            <td>✅</td>
         </tr>
         <tr>
             <td><strong>Push Notifications</strong></td>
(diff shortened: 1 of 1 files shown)


katopz:
also dashboard? http://blog.parse.com/announcements/introducing-the-parse-server-dashboard/

swarthout:
@katopz Also true. I will add that to the PR

daffl:
Thank you! I like that those comparisons keep us up to date with what's going on :smile:
We probably want to update the written comparison at https://github.com/feathersjs/feathers-docs/blob/master/why/vs/parse.md, too.

why/vs/readme.md
```diff
@@ -76,7 +76,7 @@ If you find something invalid or out of date in the comparisons, please [create
             <td>❌ 🤓</td>
             <td>❌</td>
             <td>❌</td>
-            <td>❌</td>
+            <td>✅</td>
```
ekryski: mmm sort of. You can get things to get notified about things over sockets but you can't send data over sockets. I feel like it's more of half check.

Thanks for the PR though.

ekryski: My bad. I'm an idiot. Was looking at the wrong column!

PR #97, opened 2016-03-23
Fix documentation for $select

daffl:
The documentation was incorrect (probably because it aws copied from the original NeDB and MongoDB behaviour).

Closes https://github.com/feathersjs/feathers-mongoose/issues/71


ekryski:
:shipit:

PR #104, opened 2016-03-26
Promoting the use of the CLI

derek-watson:


marshallswain:
Looks good.  Let's make it happen.

marshallswain:
I've been through the guide with this and it worked great.  This is the only issue I identified, which isn't a biggie: https://github.com/feathersjs/feathers-docs/issues/108

daffl:
I'd also like to get it in as soon as possible but maybe we can chat some more tomorrow to figure out what we want to do for releasing this. I'd mainly like to make sure that the CLI gets the attention it deserves e.g. with a separate release post on the blog vs. just quietly merging it into the docs.

ekryski:
I think we can :shipit:. We can do a quick announcement in slack and then I'll do a recap post within the next week or two on Medium that talks about all the new stuff.

PR #106, opened 2016-03-27
Update the docs to use the feathers-cli

marshallswain:
This replaces every instance of `yo` with the matching `feathers` command.


marshallswain:
Looks like a better version of this already exists in #104

PR #110, opened 2016-03-28
Fix the react placeholder.

marshallswain:
This will show a random image from this url: https://placeimg.com/60/60/people.  It will change every time the page is loaded, but I think it's better than a broken image.


daffl:
Oops. I totally thought I was using http://placehold.it URLs. I like the people even more though. Can you change this for the jQuery guide as well?

marshallswain:
Done

PR #113, opened 2016-03-29
Remove encrypt/decrypt in place of sign

gavingmiller:
JWT is not encrypting/decrypting the payload, only validating whether it has been tampered with or not.


ekryski:
Totally right. Thanks @gavingmiller! :smile:

PR #115, opened 2016-03-31
default token algorithm is HS256

farwayer:


ekryski:
Yup my bad. I had the intention of updating it to 512 but decided against it and forgot to change the docs. Thank you!

PR #116, opened 2016-03-31
Fix typo in philosophy page

Glavin001:


ekryski:
Woot! Thanks @Glavin001!

PR #119, opened 2016-04-02
Adding RethinkDB

marshallswain:


daffl:
We probably should add it to `SUMMARY.md` as well.

PR #124, opened 2016-04-03
Fixing wrong method name in sample code

erdemorman:
Sample code for `restrictToRoles` uses wrong method name (i.e. `restrictedToOwner`) instead of `restrictToRoles`.


marshallswain:
Thank you!

PR #125, opened 2016-04-06
Change feathers-client to npmcdn.com with version range

### LINKED ISSUE #120: Update feathers-client CDN links to the latest version
We may want to investigate how to automate this more easily, possibly through https://npmcdn.com/ instead of RawGit

daffl:
This switches the docs to use feathers-client from npmcdn.com. This is a great change because it supports SemVer version ranges (like https://npmcdn.com/feathers-client@^1.0.0/dist/feathers.js) and we don't need to check in the distributables anymore.

@ekryski Can we change this on the homepage Gists as well?

Closes #120

### DIFF
--- a/authentication/client.md
+++ b/authentication/client.md
@@ -43,7 +43,7 @@ Then you pass whatever other fields you need to send to authenticate. See below 
 
 ```html
 <script src="//code.jquery.com/jquery-1.12.0.min.js"></script>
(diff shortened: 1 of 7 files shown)


ekryski:
Done! :shipit:

daffl:
I think you forgot the `^` (needs to be `<script type="text/javascript" src="//npmcdn.com/feathers-client@^1.1.0/dist/feathers.js"></script>`). That way we don't have to update it with every release.

ekryski:
Fixed. Thanks @daffl

PR #126, opened 2016-04-06
Also mention 'patch' in services introduction.

juodumas:


marshallswain:
@juodumas, Thank you!

PR #127, opened 2016-04-06
Too much ES6? Fix hook init examples.

juodumas:


marshallswain:
Thanks for looking at this.  Does the existing ES6 code not work with the latest version of Node?  I think that would be the only reason we would want to go back to ES5 with any of the docs.  If it's not supported, we'll want to merge this, imo.  Would you mind checking?

juodumas:
My bad, sorry. Works fine with node-5.10.0. I thought I have tested, but apparently my testing was crooked :)

PR #133, opened 2016-04-13
Adding note and script tags to include ES6 polyfills

### LINKED ISSUE #123: Add a note to use an ES6 shim on the client
We are already showing to include core.js on the homepage. Need to double check that this is also mentioned in the client side Feathers documentation.

daffl:
Closes #123


ekryski:
:shipit:

PR #135, opened 2016-04-14
Update philosophy.md

BrainMaestro:
`POST` was listed twice. And `PUT` is usually used for updates. Is this correct? or was the original way intended.


daffl:
Indeed! Thanks!

PR #137, opened 2016-04-14
update primus link on meteor vs page

chrisfishwood:
Fix primus link typo


daffl:
We definitely suck at linking 😉 Thank you!

PR #142, opened 2016-04-16
Tutorial fix to show avatar image (React section)

pdrummond:
When I copied the source code from the react section in the docs, the Gravatar image in the chat app always showed the placeholder, even for signed-in users.  This is because the `avatar` field is used when it should be the `image` field.


mikewink:
The same is true for the jQuery version, so I guess it's better to update the gravatar. It's also the actual key in the repos for the feathers-chat example.

daffl:
Yes, I think `avatar` is correct since it is also in the chat demo app. I think we changed it because it was `avatar` already for the React native chat. I merged #143 and going to close this one and make sure that the other frontend guides are updated as well. Thanks for bringing that up!

PR #144, opened 2016-04-17
Update user-management.md

tryy3:
Noticed that there were no space between the type and name tag.
It wont effect the actual page or anything, but if you are reading line by line on the docs, the syntax will be screwed up.
https://gyazo.com/c19d9f54ab2f064da138e28931d556d3 as seen in the picture, you can see that githubs generator can't detect that there is a new tag there.

### DIFF
--- a/getting-started/user-management.md
+++ b/getting-started/user-management.md
@@ -25,7 +25,7 @@ Let's replace `public/index.html` with the following welcome page:
   <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8">
     <meta name="viewport"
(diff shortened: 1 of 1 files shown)


tryy3:
oh I also noticed that the only tag that has a ending slash / is the 
`    <meta name="viewport"
      content="width=device-width, initial-scale=1.0, maximum-scale=1, user-scalable=0" />`
tag.

I am new to open source and I am not sure if adding a slash actually does any impact on the code.
But it looked a bit funny when there are several tags that doesn't have a closing tag, but thats the only tag that actually ends with a slash.

I could fix those "typos" too, so there is a bigger reason for this pull request.

daffl:
Perfect. Thank you.

PR #156, opened 2016-04-24
moving environment setup section to app tutorial

ekryski:
I think this gives better visibility. Plus, I think because the guide wasn't in the `SUMMARY.md` file the previous links to the guide were broken.


ekryski:
Disregard the big changes in SUMMARY.md. Someone messed with the indentation at some point and I'm a bit OCD so I **had** to fix it.

Issue #155, opened 2016-04-24
Write a guide for creating plugins


ekryski:

thejones:
I started on this & there should be a Pull Request (https://github.com/feathersjs/feathers-docs/pull/160) . I have no idea what I am talking about so feel free to edit/make changes.

thejones:
@daffl do you want to keep this open for your modifications to the guide?

daffl:
I'll make those some time later but I think we can totally close this one. Thanks a bunch again!

ekryski:
Yeah! Thanks @thejones! Nice work!

PR #160, opened 2016-04-29
Basic plugin creation doc

thejones:
References issue https://github.com/feathersjs/feathers-docs/issues/155

### DIFF
--- a/guides/creating-a-plugin.md
+++ b/guides/creating-a-plugin.md
@@ -31,7 +31,7 @@ Output files from generator:
    create .npmignore
 ```
 
-Simple right? We technically could call it a day as we have created a Plugin. However, we probably want to do just a bit more. Generally speaking a Plugin is a [Service](../services/readme.md). The fun part is that a Plugin can contain multiple Services which we will see below. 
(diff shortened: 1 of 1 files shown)


guides/creating-a-plugin.md
```diff
@@ -1,3 +1,318 @@
...
+   create src/index.js
+   create test/index.test.js
+   create README.md
+   create LICENSE
+   create .gitignore
+   create .npmignore
+```
+
+Simple right? We technically could call it a day as we have created a Plugin. However, we probably want to do just a bit more. Generally speaking a Plugin is a [Service](../services/readme.md). The fun part is that a Plugin can contain multiple Services which we will see below. 
+
```
marshallswain: It would be nice at this point to explain what the tutorial services will be, so readers have an idea of the functionality that they will be building.

marshallswain:
@jones, this is really awesome!

guides/creating-a-plugin.md
thejones: Added a bit to this section.

daffl:
This is great, thank you! I'm going to merge this tomorrow. I hope it's ok if I change it up a little bit after that. I'd like to add some information on how to create your own spec compliant service adapter and maybe also mention that a plugin does not have to be a service but could also be used to mix functionality into services (like what `feathers-hooks` does) and the application.

thejones:
Rock on.

PR #161, opened 2016-04-30
Missing to app.service call

shadowRR:
Getting the app.service for `userService` and `messageService` are missing.


daffl:
I think this is absolutely correct but I'll let @niallobrien double check.

niallobrien:
Good catch, thanks for the PR. :)

shadowRR:
You're welcome. Great tutorial by the way, thanks a lot for putting it together ;)

PR #162, opened 2016-05-02
fixed typo and removed extra spaces in vuejs framework section

quentin-sommer:


daffl:
Most of Feathers contributors would argue that _initialise_ is the correct spelling but we're not sticklers about it 😉  Thank you for the PR!

quentin-sommer:
It stroke me since I'm French and we write it initialise :p

PR #163, opened 2016-05-04
Fixed broken Services link in API chapter

MitMaro:
Not sure how to test this correctly, but I tried to follow existing links I found in other chapters.


marshallswain:
@MitMaro Thanks!

PR #168, opened 2016-05-12
Update primus.md

aminmeyghani:
It seems like `hooks()` should be passed into `configure` instead of `configure(feathers.hooks())`...


daffl:
Good catch! Looks like some 1.0 docs snuck in there.

PR #170, opened 2016-05-17
Update client snippets

silvestreh:
So they `require` `feathers/client` rather than `feathers`


marshallswain:
Oh good catch.

marshallswain:
Thank you!

PR #177, opened 2016-06-01
Added test command that works with ES2015 (Babel)

harunsmrkovic:
Referenced command from [https://babeljs.io/docs/setup/](https://babeljs.io/docs/setup/#installation)


daffl:
Makes sense. Thank you!

PR #182, opened 2016-06-06
Updated to docs to match hook generated error message - fixes #181

### LINKED ISSUE #181: Documentation is incorrect for feathers-authentication bundled hook 'restrictToRoles'
https://github.com/feathersjs/feathers-docs/blob/master/authorization/bundled-hooks.md#L159

Documentation states that `restrictToRoles` can be used on all request types when the error message says as follows:

```Javascript
{
    "name": "MethodNotAllowed",
    "message": "The 'restrictToRoles' hook should only be used on the 'get', 'update', 'patch' and 'remove' service methods if you are using the 'owner' field.",
    "code": 405,
    "className": "method-not-allowed",
    "errors": {}
}
```

Mentioum:


daffl:
👍

PR #185, opened 2016-06-09
getTime is not a constructor

obouchari:
In the code examples `new Date.getTime()` should be `(new Date()).getTime()`

### DIFF
--- a/services/readme.md
+++ b/services/readme.md
@@ -65,7 +65,7 @@ app.use('/messages', {
       id,
       read: false,
       text: `Feathers is great!`,
-      createdAt: new Date().getTime()
+      createdAt: (new Date()).getTime()
     });
   }
 });


daffl:
What platform is that happening on? If I run `new Date().getTime()` in the console or the REPL it does what I expect.

marshallswain:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence
According to operator precedence, this should work just fine.  It works in all browser environments.  If we were doing `new Date.getTime()` then it would be different, since the `.` operator has higher precedence, but when you pass arguments `()`, even if it's empty parentheses, the constructor is handled with the first function.

obouchari:
In this case the parenthesis are messing.  in the examples it's `new Date.getTime()` instead of `new Date().getTime()`. BTW I love this framework so far I just discovered it yesterday and I already like it its easy to understand and extend. Thanks.

PR #186, opened 2016-06-12
Update faq.md

chrissm79:
Couldn't populate the related documents unless I added the query param key.


daffl:
That was totally still the old client syntax. Thank you!

PR #209, opened 2016-08-05
Change syntax highlighting to javascript

kulakowka:
For a better display in gitbook


daffl:
Does it support JSX now? I think I went back and forth for a bit because it made it look like syntax errors. Let's give it a try.

kulakowka:
I think it would be correct.

PR #223, opened 2016-09-13
Updated the section related to service timeout

kaiquewdev:
Exaplanation included to the doc for a better example about timeout usage.


daffl:
Awesome, thank you!

PR #228, opened 2016-09-28
Fix "io" import statement

Audiotuete:
Added `/socket.io` to `import io from 'socket.io-client';`

### DIFF
--- a/clients/feathers.md
+++ b/clients/feathers.md
@@ -66,7 +66,7 @@ Then in the main application file:
 ```js
 import 'babel-polyfill';
 import './user-agent';
-import io from 'socket.io-client';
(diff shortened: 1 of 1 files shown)


kulakowka:
Could you explain why we should do so?

Audiotuete:
Sure. Sorry, wrote it in the slack channel so how should anyone else know. It's because without it throws `Requiring unknown module "xmlhttprequest-ssl`. ekryski already changed it on the landing page example, expect he uses the `const io = require('socket.io-client/socket.io');` since the polyfill and bugfix are handled differently there.

daffl:
Cool. Thank you and thank you for the explanation.

PR #229, opened 2016-09-28
Minor editorial tweaks to include $select in header

sjmcdowall:
Just changed the header and intro paragraph to mention selecting so people can find it easier ..


kulakowka:
@sjmcdowall many thanks!

PR #233, opened 2016-10-07
Added docs for feathers-hooks-common

eddyystop:
Please check out the format, etc. Comments are welcome.

I noticed someplace that someone posted a todo about documenting how to validate with hooks. I added a validation.md because I felt an example was needed to properly document the value of the validate hook.

### DIFF
--- a/hooks/common_hooks.md
+++ b/hooks/common_hooks.md
@@ -1,0 +1,645 @@
+# Common hooks
+
+When it makes sense to do so, some plug-ins include their own hooks. The following plug-ins come bundled with useful hooks:
+
+- `feathers-hooks` (see note below)
+- [`feathers-mongoose`](../databases/mongoose.md)
+- [`feathers-authentication`](../authorization/bundled-hooks.md)
+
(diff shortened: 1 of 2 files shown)


daffl:
Cool, looks great. One question: Should we remove the inline JSDoc documentation from the source? I see challenges with keeping both, this document and the inline docs in sync (and they basically contain the same thing right?).

eddyystop:
Yea, I understand your point. The inline, the README and GitBook are basically the same thing and  I would deprecate the README right away.

These days I tend to like the inline as its close to the code and easy to change when you change the code, also easier to check that its correct. I look at the others as derived from the inline.

This opinion could change with a lot of doc to support. I wouldn't mind watching PRs and keeping the inline and GitBook sync'ed for hooks-common. 

As an aside, you can avoid the JSDoc liniting by placing a blank line between the comments and the function. So that shouldn't get in anyone's way. Eh?

daffl:
Allright, merged. Could you add it to the table of contents at `summary.md` as well?

eddyystop:
Do you want Hook / Built-in hooks to remain?

PR #234, opened 2016-10-07
Tighten-up file uploading section, goals subsection.

StevenBlack:
The *Our Goals* section rang somwhat oddly, with a mix of present and future verb tenses.  Also some “filler” words could be removed without affecting clarity.

So here the *Our Goals* section is a little tighter, and a little more consistent.


daffl:
👍

PR #239, opened 2016-10-10
Added: Socket Debug Hints

psi-4ward:


kulakowka:
@psi-4ward thank you!

PR #240, opened 2016-10-11
More (simple) examples of using built-in operands

sjmcdowall:
After being totally confused as how to use operands like $le, $gt, $ne, etc. within a standard URL type GET request, decided to add a couple of examples to give people a basic idea.

### DIFF
--- a/clients/vanilla-rest.md
+++ b/clients/vanilla-rest.md
@@ -18,6 +18,18 @@ GET /messages?status=read&user=10
 
 Will call `messages.find({ query: { status: 'read', user: '10' } })` on the server.
 
+If you want to use any of the built-in find operands ($le, $lt, $ne, $eq, $in, etc.) the general format is as follows:
+
+```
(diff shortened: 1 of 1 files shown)


kulakowka:
@sjmcdowall thank you! It's a nice additional.

sjmcdowall:
I'll do some more once I actually get anything else working.  Can't get $in to work with the feathers-rethinkdb adaptor for anything. :(

kulakowka:
@sjmcdowall 
> After more than seven years of development, the company behind [RethinkDB is shutting down](https://rethinkdb.com/blog/rethinkdb-shutdown/). 

It seems the fate of Rethinkdb not clear.

sjmcdowall:
Yep -- quite aware of the saga.  They appear to be sending it to open-source land .. and who knows... they were basically absorbed by Stripe .. so maybe all is not lost.  <shrug> ...

PR #241, opened 2016-10-12
More examples of REST Calls

sjmcdowall:
Just expanded the documentation with some more (slightly) complex style FIND queries and introduced here the qs() option to make even more complex queries possible via the URL.

### DIFF
--- a/clients/vanilla-rest.md
+++ b/clients/vanilla-rest.md
@@ -30,6 +30,29 @@ For example, to find the records where field _status_ is not equal to **active**
 GET /messages?status[$ne]=active
 ```
 
+To get really complex queries using the URL syntax, the back-end must implement the extended query string qs() (query string parser) middleware layer.  This is accomplished within the code by adding a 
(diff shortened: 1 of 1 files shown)


clients/vanilla-rest.md
```diff
@@ -30,6 +30,29 @@ For example, to find the records where field _status_ is not equal to **active**
 GET /messages?status[$ne]=active
 ```
 
+To get really complex queries using the URL syntax, the back-end must implement the extended query string qs() (query string parser) middleware layer.  This is accomplished within the code by adding a 
+```
+app.set('query parser', 'extended');
```
marshallswain: Does it make any difference doing what you've written here or doing the following?

```
.use(bodyParser.urlencoded({ extended: true }))
```

marshallswain: ahh.  thanks.

daffl:
Great pointers, thank you!

PR #244, opened 2016-10-12
Added: paginate:false only for server hint

psi-4ward:


daffl:
Good hint, thank you!

PR #245, opened 2016-10-12
Fix the path to user-model.js

aurerua:
Path before editing: src/services/menu/user-model.js
Path now: src/services/user/user-model.js


aurerua:
You're welcome! Thanks for Feathers :)

PR #249, opened 2016-10-16
Minor changes to GitBook

eddyystop:
- Added common hooks to summary.md
- Added deprecated notice added to bundled.md
- Added live-query example to showcase.md
- Minor change in showcase.md to force 2 lines which don't appear
in GitBook.

### DIFF
--- a/hooks/bundled.md
+++ b/hooks/bundled.md
@@ -1,6 +1,6 @@
 # Built-in hooks
 
-| The built-in hooks are being deprecated in favor of [Common Hooks](./common_hooks.md). Please start using Common Hooks.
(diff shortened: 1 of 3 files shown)


hooks/bundled.md
```diff
@@ -1,5 +1,7 @@
 # Built-in hooks
 
+| The built-in hooks are being deprecated in favor of [Common Hooks](./common_hooks.md). Please start using Common Hooks.
```
ekryski: maybe we should highlight this like a ProTip or make it bold so it stands out more.

ekryski:
:shipit:

PR #255, opened 2016-10-21
Get feathers-hooks-common doc showing in GitBook TOC.

eddyystop:
Cut hooks/common.md in half to see if the original
was too large for GitBook.

### DIFF
--- a/hooks/common-complete.md
+++ b/hooks/common-complete.md
@@ -325,3 +325,321 @@ app.service('stockItems').before({
 #### Options
 
 - `fieldName` [optional. default: `deleted`] - The name of the field holding the deleted flag.
+
+## Utilities
+
+### setSlug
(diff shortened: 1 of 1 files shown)


ekryski:
:shipit:

ekryski:
Actually @eddyystop it looks like you need to add them to the `SUMMARY.md` file. So either call this file the same thing `hooks/bundled.md` or update the SUMMARY.md file to have `* [Built-in hooks](hooks/bundled-complete.md)`

eddyystop:
@ekryski  I was trying to see if the original common_hooks.md was too large. So I had SUMMARY.md pointing to a smaller version.

DO NOT MERGE THIS PR. @ekryski may have fixed the problem we were having with GitBook. I'll delete PR once we see the problem is gone.

daffl:
I hope it's ok if I close this one. GitBook still seems to be broken but it'll hopefully just rebuild once we fix it.

PR #264, opened 2016-10-28
Clarify when a service's setup method is called.

edwardmsmith:
I spent a bit of time trying to figure out why my service's `setup` method was not being called (in testing, and in interactive node.js sessions) when the service was registered.  Going into feather code, I realized that the service's `setup` is only called once `app.listen` is invoked.  I thought this clarification might be useful to others.


daffl:
Thank you. It is already mentioned in https://docs.feathersjs.com/api/readme.html#setup but it probably makes sense to do that again, here.

PR #265, opened 2016-10-31
Hooks update

ekryski:
Update and consolidations of hook docs. We should tag this release.


eddyystop:
`common.md` is 621 non-blank lines now. It will grow with a new big-populate hook, and with other enhancements accumulating in the issues.

Should we break some of the larger sections into their own files?

This should not stop this PR being merged though. I'm looking forward to proof-reading the results.

ekryski:
@eddyystop I was thinking that too. I think let's address chopping things up when we overhaul the docs. Likely coming up soon once auth and the generator are done.

daffl:
We have to bring the built-in hooks page back.

ekryski:
@daffl all the hooks are in `common.md` now or did I miss some?

PR #266, opened 2016-11-01
adding bundled hooks back for now

ekryski:


daffl:
👍  Thank you for adding those back.

PR #267, opened 2016-11-02
Improved common hooks doc

eddyystop:
- Validations section is very large. Removed text that offered little.
- Added anchors so the TOC links at top of doc work. Its very awkward
navigating this doc with links.
- Remove emphasis from Options subsections as they made it
visually harder to see the headers for each hook.
- Some text made more terse.


ekryski:
:shipit: Does bolding the "options" look any better?

ekryski:
I agree, we'll split these out soon. I think @daffl is putting together a high level outline for the new doc structure so we can all discuss on Friday.

eddyystop:
I think the options stand out well enough as text because of their surroundings.

PR #271, opened 2016-11-02
resetting inline links back to default style

ekryski:


ekryski:
Gtihub webhooks seem a little slow right now. Will push to master to see if this kicks off.

PR #272, opened 2016-11-03
Publish to Surge.sh with TravisCI

marshallswain:


daffl:
Cool. I know we still have to get it passing but other question: Does this also build the eBook and PDF versions? If not would it be hard to add those as well?

marshallswain:
I haven't looked, yet, but I doubt it is hard.  For now the Gitbook CI is hopefully producing a real PDF and not a blank one.

marshallswain:
It's working, so we at least have that in the meanwhile.

marshallswain:
Cool.  Node 6.5 to the rescue!

PR #273, opened 2016-11-05
added missing explanation on activating gravatar hook in user service

martinhecher:


daffl:
Thank you but `feathers generate hook` should already take care of that automatically. Did that not happen for you?

PR #286, opened 2016-11-14
Fix RethinkDB documentation example

luke3butler:
Create 'messages' as a table instead of a database


daffl:
Thank you! Will make it into the documentation updates I'm working on for the next adapter release.

PR #289, opened 2016-11-15
[Events] Add 'Event Listener Lifecycle' section

petermikitsh:
I ran into a situation in React where I had to register and unregister an event listener. I was familiar with the registering api (`service.on(...)`), but I didn't know about `service.removeListener(...)` until I got some help from @ekryski.

I figured it'd probably be good to have this in the docs.

### DIFF
--- a/real-time/events.md
+++ b/real-time/events.md
@@ -125,6 +125,20 @@ Now clients can listen to the `<servicepath> status` event. Custom events can be
 
 It is easy to listen for these events on the client or the server. Depending on the socket library you are using it is a bit different so refer to either the [Socket.io](socket-io.md) or [Primus](primus.md) docs.
 
+## Event Listener Lifecycle
+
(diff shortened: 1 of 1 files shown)


daffl:
The top of [this page](https://docs.feathersjs.com/real-time/events.html) says:

> Once registered through app.use, a Feathers service gets turned into an EventEmitter that 

With a link to [Nodes EventEmitter](https://nodejs.org/api/events.html). Maybe we want to list a short overview of the entire API if it is easy to miss.

petermikitsh:
Yeah, that sounds like a good idea. It never occurred to me to click that link.

To be honest, I didn't think it would even be possible to unregister a hook when I posted my question in Slack since I only saw code samples showing registering listeners and no examples of their removal.

ekryski:
I think we should :shipit:. It's totally useful for people to see how you would unregister for events and this is short and sweet.

PR #294, opened 2016-11-20
Use correct ico file

miketheman:
The image is not included in the generated template, and returns a 404.

It is also now consistent with the other html files.


ekryski:
Thanks @miketheman!

Issue #322, opened 2016-12-01
Wrong package name in docs


jinwyp:
https://docs.feathersjs.com/hooks/common.html

ProTip: Hooks previously in feathers-hooks are included with the common hooks. As of v1.6.0, feathers-hooks instead exports **feathers-hooks-common** to provide backward compatibility. feathers-hooks will become part of core in Feathers 3.0 and you will have to import **feathers-hooks-common** separately.

package name "feathers-hooks-common" should be "feathers-common-hooks" .  https://github.com/feathersjs/feathers-hooks-common

when I type npm i feathers-hooks-common get error .

eddyystop:
`feathers-hooks-common` is the repo being used, and so the name in the docs is correct.

I did not experience any problems running `npm i fedathers-hooks-common`.

What error did you get?
What version of node (`node -v`) and npm (`npm -v`) are you running?

jinwyp:
Sorry , I wrote wrong link. 

the wrong text on page https://docs.feathersjs.com/hooks/bundled-deprecated.html 

ProTip: The built-in hooks are being deprecated in favor of Common Hooks. All of these are currently available as part of **feathers-common-hooks**. Please start using that module.

package name "feathers-common-hooks" should be "feathers-hooks-common" .

when I type npm i feathers-common-hooks get error .

daffl:
Closed via #325

PR #325, opened 2016-12-02
Wrong package name "feathers-hooks-common" in docs 

jinwyp:
https://github.com/feathersjs/feathers-docs/issues/322


daffl:
Thank you! Could you remove the .gitignore updates (or at least the user specific stuff. If we are ignoring the `.idea` folder already there shouldn't be a need to add individual files in that folder).

PR #327, opened 2016-12-02
Replace all SVG links with PNGs so that GitBook PDF and eBook build w…

daffl:
…orks again. They're definitely not that pretty but at least the ebook versions should work.


marshallswain:
Oh you found them.  :+1:  You get 💯 points.

daffl:
It's too bad though, the shields did look really nice.

PR #336, opened 2016-12-09
Added sections to conditional hooks page

eddyystop:
- 5 new hooks have to be documented
- The new sectioning makes it easy to see what the hooks are for
- Placeholders opnly added for when, unless, some, every


eddyystop:
ekryski wrote"Yup looks good! Go for it!"

PR #339, opened 2016-12-09
Fixed docs for iffElse, iff, isNot

eddyystop:
- They all allow a predicate that is a boolean, Promise or func
- Docs only mentioned func


ekryski:
Looks good. :shipit:

PR #340, opened 2016-12-12
Typo fix in setUpdatedAt hook

shadowRR:
Small typo when using the `setUpdatedAt` hook, which was shown as `setCreatedAt`.


eddyystop:
Thanks!

PR #355, opened 2016-12-28
Update docs.feathersjs.com/services/readme.html

crobinson42:


eddyystop:
Thanks.

However that should be `../hooks/bundled-deprecated.md#disable`. Would you mind making the change?

PR #356, opened 2016-12-29
Fixed description of verifyChanges field in authentication management

chrisbujok:
Just a small change, but I lost a lot of time because I've set `verifyChanges` field type to array of strings in my Sequelize/Postgres model and it didn't work at all. After looking over the source code I realized it does not use array of strings, but a key-value map. Setting column type to JSON solved the problem, but I think that description in docs is confusing, so I decided to create this PR.


eddyystop:
Thanks for the catch!

PR #362, opened 2017-01-06
Socket.io-client dependency import

gor181:
Change `import io from 'socket.io-client/socket.io';` to `import io from 'socket.io-client';` as otherwise there's an error when importing the dependency. (At least from what I have been encountering when switching our app from rest client to socket client).

It's used as described in [official repo](https://github.com/socketio/socket.io-client#nodejs-server-side-usage) too.

### DIFF
--- a/clients/feathers.md
+++ b/clients/feathers.md
@@ -59,7 +59,7 @@ Then in the main application file:
 
 ```js
 import 'babel-polyfill';
-import io from 'socket.io-client/socket.io';
+import io from 'socket.io-client';
 import feathers from 'feathers/client';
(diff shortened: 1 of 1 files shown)


marshallswain:
It should probably be

```js
import io from 'socket.io-client/dist/socket.io';
```

since the dist version is compatible with more module loaders.

daffl:
I'm not sure. Steal seems to be the only module loader that doesn't work with the `socket.io-client` library directly.

marshallswain:
Fair enough @daffl.  Merged.   Thanks for the contribution, @gor181!

PR #368, opened 2017-01-09
Update angular.md

roganmelo:
The documentation shows feathers-client using angular, through angular service creation that encapsulates the feathers-client configuration and an example of use in an angular controller.


daffl:
Thank you again for the PR. I am going to close all Angular guide related PRs since we don't have anybody who can review them and we are about to remove frontend framework guides from the documentation and move them into blog posts.

We are still looking for anybody interested in helping with a chat guide blog post for Angular 2.

PR #370, opened 2017-01-10
Rename service from 'messages' to 'message'

sambostock:
The scaffolding section of the tutorial expects the service to be named 'message', not 'messages'.

### DIFF
--- a/getting-started/quick-start.md
+++ b/getting-started/quick-start.md
@@ -33,7 +33,7 @@ With [NodeJS](https://nodejs.org) installed, you can quickly scaffold your first
     $ feathers generate
     ```
 
-4. Generate a new service. When asked `What do you want to call your service?`, type `messages` and confirm the other prompts with the default:
+4. Generate a new service. When asked `What do you want to call your service?`, type `message` and confirm the other prompts with the default:
 
     ```
     $ feathers generate service


eddyystop:
@daffl. Are we sure about this?

Authentication creates a folder for `user` not `users` though the path is `/users`.

It may be common that people go `users = app.service('/users')` but the generator goes `userService = app.service('/users')`.

daffl:
We should keep it for now because the generator automatically adds pluralization (unless this broke for you @sambostock). This will be fixed in the next generator version (related issue: https://github.com/feathersjs/generator-feathers/issues/44)

sambostock:
My main issue is that [quick-start.md](https://github.com/feathersjs/feathers-docs/blob/cc11d7dfa3e0adbea3b046a8c40a416630b84dcb/getting-started/quick-start.md) instructs the learner to generate a `messages` service

```
4. Generate a new service. When asked `What do you want to call your service?`, type `messages` and confirm the other prompts with the default:

    ```
    $ feathers generate service
    ```
```

while [scaffolding.md](https://github.com/feathersjs/feathers-docs/blob/b101cb5eab50286ae54cf3041312a7358ee77420/getting-started/scaffolding.md) instructs the learner to generate a `message` service:

````
```
$ generate service
```

The name should be `message` and the other options can be accepted with their default value (we will add authorization in the next chapter):
````

Despite that, [scaffolding.md](https://github.com/feathersjs/feathers-docs/blob/b101cb5eab50286ae54cf3041312a7358ee77420/getting-started/scaffolding.md) begins by stating:

> If you've already gone through the [Quick Start](https://github.com/feathersjs/feathers-docs/blob/cc11d7dfa3e0adbea3b046a8c40a416630b84dcb/getting-started/quick-start.md) section you can skip ahead to [What Just Happened?](https://github.com/feathersjs/feathers-docs/blob/b101cb5eab50286ae54cf3041312a7358ee77420/getting-started/scaffolding.md#what-just-happened).

Following those instructions causes the tutorial to break once the learner gets to the [Processing Messages section](https://github.com/feathersjs/feathers-docs/blob/b101cb5eab50286ae54cf3041312a7358ee77420/getting-started/formatting-and-hooks.html#processing-messages), which expects the service (and the relevant source file paths) to be named `message`.

daffl:
Thank you again for the PR. The naming scheme (and several other things) will be updated in the new guide (see #442)

PR #371, opened 2017-01-11
Don't use `const` for non-constants

IBwWG:


daffl:
But the object is a constant and shouldn't be reassigned no? To avoid modifying an object we'd use `Object.freeze`.

IBwWG:
Thanks @daffl , you're totally right.  "It does not mean the value it holds is immutable, just that the variable identifier cannot be reassigned. For instance, in case the content is an object, this means the object itself can still be altered."

PR #383, opened 2017-01-27
Minor change in Angular2 documentation

emin93:
Now using feathers-client npm module since this one has everything included with type definitions for FeathersApp and FeathersService

### DIFF
--- a/frameworks/angular2.md
+++ b/frameworks/angular2.md
@@ -13,14 +13,14 @@ We'll be setting up the basic service structure for an application that uses `fe
 If you're using TypeScript to develop with Angular 2, Feathers and its related modules should be loaded into your app through npm packages. `npm install` the ones you'll need:
 
 ```bash
-npm install feathers feathers-hooks feathers-rest feathers-socketio socket.io-client feathers-authentication
+npm install feathers-client --save
 ```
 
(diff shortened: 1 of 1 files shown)


daffl:
Thank you but I don't think this is correct. The code does use the separate client modules (e.g. https://github.com/feathersjs/feathers-docs/pull/383/files#diff-19f9973ab295886e5b3db9ab4d1c19f8R25) not the different bundled feathers-client module.

emin93:
At first i tried it with the documented modules, but those didn't have any typescript definition files. So i started researching and stumbled upon this module bundle, which seemed to contain all needed modules including type definitions. Correct me if i'm wrong.

With just feathers-client in the package.json, the following modules got installed:
bildschirmfoto 2017-02-15 um 08 40 38

emin93:
In the Angular2 doc you recommend to install the module "feathers", which also contains non-client related files. Imho this is not the expected behavior. When installing "feathers-client", you only get client related files and that is correct.

daffl:
Thank you again for the PR. I am going to close all Angular guide related PRs since we don't have anybody who can review them and we are about to remove frontend framework guides from the documentation and move them into blog posts.

We are still looking for anybody interested in helping with a chat guide blog post for Angular 2.

PR #385, opened 2017-01-28
feathers.hooks(), not simply hooks()

shawnlauzon:


marshallswain:
Thanks @shawnlauzon.  I think in order for that to work we need to change the import of `feathers` to `feathers-client`.  Either that or `import hooks from feathers-hooks` and leave the variable as hooks.

daffl:
Thank you @shawnlauzon. I agree with @marshallswain I think that code sample is simply missing a

```
const hooks = require('feathers-hooks');
```

at the top instead. Would you mind updating your PR accordingly?

shawnlauzon:
Thank you, you're right of course. I'll update.

shawnlauzon:
Ok, fixed as well as some other minor changes. My code now works, thanks :)

daffl:
Thank you!

PR #388, opened 2017-02-02
Adding guides to see what TOC would look like

eddyystop:


daffl:
Let's get it in and see how it looks.

PR #397, opened 2017-02-20
Cleaning up help section

ekryski:
- removing changelog from menu (the Feathers core changelog doesn't really make sense here).
- Moving contributing, faq, and help top level and into their own folders


ekryski:
A bunch of the links in the FAQ section are broken now so not sure what to do about those since the content they used to link to has changed significantly. Frankly most of them are better suited for guides.

cc/ @daffl thoughts??

PR #401, opened 2017-02-21
Removed socket.io from supported in primus

datoml:
Regarding to https://github.com/primus/primus#supported-real-time-frameworks Primus doesn't support socket.io anymore.


bertho-zero:
https://github.com/primus/primus#engineio

PR #411, opened 2017-03-01
improve docs based on added typescript support

superbarne:
feathersjs/feathers#381 angular2 docs


daffl:
Thank you for being so thorough!

PR #427, opened 2017-03-18
Make ecosystem plugins more accessible

marshallswain:
This makes the Ecosystem plugins more accessible by bringing them into the left sidebar.  The list is huge, so this is very helpful to see how amazing our community is.  This is what it looks like:

screen shot 2017-03-18 at 7 55 58 am
 
Ideally, we could bold the headings that I've artificially bolded in the screenshot.  @eddyystop or @daffl is there a way to control this with the plugin you've made for the guides?


ekryski:
@marshallswain did you figure out whether you can bold the headings like that? I think you can, and I like that look.

PR #439, opened 2017-03-20
Fix remove method description

mgesmundo:


daffl:
Thank you for the pull request but I don't think that would work. If the `id` is null, `params.query` has to be a query matching the items you want to remove.

mgesmundo:
O yes of course. Can I (or you) update the documentation to explain this?

daffl:
This has been done in the new documentation branch at https://docs.feathersjs.com/v/auk-reorg/api/services.html#removeid-params which will be published shortly.

PR #457, opened 2017-04-05
timestampData: true

bugss:
add timestampData:true . avoid sometime update will exception with a before update hook like hooks.setCreatedAt(), hooks.setUpdatedAt()
#59


marshallswain:
This is a good idea.  Are you sometimes seeing an exception when using the mentioned hooks?

daffl:
Agreed, it's a good change. Could you re-submit it against latest master and then close this one though?

daffl:
I'm going to close this one (see previous comment). I'd be happy to see it against the generator or latest master.

PR #458, opened 2017-04-05
Start the `populate` documentation with a simple working example

hubgit:
As the syntax for the `populate` hook is quite complex, give a simple working example at the start of the documentation.

### DIFF
--- a/api/hooks-common.md
+++ b/api/hooks-common.md
@@ -492,42 +492,26 @@ Populates items *recursively* to any depth. Supports 1:1, 1:n and n:1 relationsh
 - Backward compatible with the old FeathersJS `populate` hook.
 
 ```javascript
-const schema = {
-  service: '...',
-  permissions: '...',
-  include: [
-    {
-      service: 'users',
-      nameAs: 'authorItem',
-      parentField: 'author',
-      childField: 'id',
-      include: [ ... ],
-    },
-    {
-      service: 'comments',
-      parentField: 'id',
-      childField: 'postId',
-      query: {
-        $limit: 5,
-        $select: ['title', 'content', 'postId'],
-        $sort: {createdAt: -1}
-      },
-      select: (hook, parent, depth) => ({ $limit: 6 }),
-      asArray: true,
-    },
-    {
-      service: 'users',
-      permissions: '...',
-      nameAs: 'readers',
-      parentField: 'readers',
-      childField: 'id'
-    }
-  ],
-};
+import { populate } from 'feathers-hooks-common';
 
-module.exports.after = {
-  all: populate({ schema, checkPermissions, profile: true })
-};
+service.hooks({
+  after: {
+    all: [
+      populate({
+        schema: {
+          include: [
+            {
+              service: 'users', // the name of the service to call `find` on
+              nameAs: 'authorUser', // the name of the property to be added to the parent
+              parentField: 'author',  // the name of the parent property that contains the child id
+              childField: '_id', // the name of the child property that contains the id
+            }
+          ]
+        }
+      })
+    ]
+  }
+});
 ```
 
 Options


eddyystop:
So instead of
```js
const schema = {
  service: '...',
  permissions: '...',
  include: [
    {
      service: 'users',
      nameAs: 'authorItem',
      parentField: 'author',
      childField: 'id',
      include: [ ... ],
    },
    {
      service: 'comments',
      parentField: 'id',
      childField: 'postId',
      query: {
        $limit: 5,
        $select: ['title', 'content', 'postId'],
        $sort: {createdAt: -1}
      },
      select: (hook, parent, depth) => ({ $limit: 6 }),
      asArray: true,
    },
    {
      service: 'users',
      permissions: '...',
      nameAs: 'readers',
      parentField: 'readers',
      childField: 'id'
    }
  ],
};

module.exports.after = {
  all: populate({ schema, checkPermissions, profile: true })
};
```

We can have
```js
const schema = {
  include: [
    {
      service: 'users',
      nameAs: 'authorItem',
      parentField: 'author',
      childField: 'id',
    }
  ],
};

module.exports.after = {
  all: populate({ schema })
};
```

And that's a major improvement?

hubgit:
@eddyystop That's not the code in the pull request, which is a fully working and commented example from which someone can copy and paste into their hooks.

Given how much trial-and-error it took to understand the current documentation, anything is a major improvement 😇

hubgit:
> The real problem is that, even for the simplest of cases, some people doesn't spend time reading even the start of the docs.

This isn't the real problem. I read the docs several times, and this one hook is particularly difficult to understand, because a) the syntax is more complex than a user might expect, b) the example code uses a different style from the other examples on the page, and c) it uses object destructuring which makes the required structure easier to miss.

eddyystop:
@hubgit Sorry anout my rant above. Once again it was not targetted at you. It comes from **hours** spent helping people who seemingly didn't try to read the docs.

In your PR
```js
parentField: 'author',  // the name of the parent property that contains the child id
childField: '_id', // the name of the child property that contains the id
``
people will still get those fields backwards. We spent several days settling down on `parentField` and `childField` as the clearest, and yet they are often coded in reverse.

I think a model of the parent and child tables would be needed.

**Comment by hubgit** [comment]
> We spent several days settling down on parentField and childField as the clearest, and yet they are often coded in reverse.

Agreed, it's difficult to think of a clearer name for those properties, and yet it's always cognitively difficult to think through which way round they should be.

**Comment by eddyystop** [comment]
Could you add some model definitions, perhaps as comments, to the PR?

**Comment by hubgit** [comment]
How about using the standard Feathers query syntax to express the query, something like this?:

```
hooks.populate({
  ownerUser: { 
    service: 'users',
    query: item => ({ id: item.owner })
  },
  authorUsers: { 
    service: 'users',
    query: item => ({ id: { $in: item.authors } })
    populate: {
      userAddress: {
        service: 'addresses',
        query: user => ({ id: { $eq: user.address } })
      }
    }
  }
})
```

eddyystop:
@Alf Eaton , the populate syntax is similar to that used by others, e.g. https://github.com/MichaelErmer/feathers-populate-hook.

Too many people are using the populate syntax to change it now, even if we wanted to. Furthermore a Sequelize-specific hook is under development using a similar syntax.

Adding examples is one thing, a rewrite is another.

hubgit:
@eddyystop That's reasonable 👍  

I've started [a new hook](https://github.com/hubgit/feathers-hook-fetch) that populates properties using standard calls to services, to see if it makes anything easier to understand.

eddyystop:
@Alf Eaton, `populate` can handle either `parentField` or `childField` (but not both) being an array as these are popular.

It'll be expanded to handle a prop name within an array of objects, as people been running into these.

You may want to support some of these.

eddyystop:
I added several examples along the lines you suggested to the new Auk version of the docs. Thanks for your perseverance. https://github.com/feathersjs/feathers-docs/pull/511

Please open a new issue if you have suggestions about these examples.

PR #502, opened 2017-04-11
Stashed some old files to see if we reuse them.

eddyystop:


daffl:
Could we put those in an old branch if we don't use them? Just want to avoid having old unused files piling up in the repo.

eddyystop:
I stashed them locally. https://github.com/feathersjs/feathers-docs/pull/504

PR #525, opened 2017-04-16
Adding effect of authentication-hooks in internal call

sasdf:


eddyystop:
What does "All hooks in this page won't affect internal call." mean?

daffl:
I will elaborate. Basically most (or all) run only if `provider` is set.

PR #530, opened 2017-04-18
Fix disableMultiItemChange example

sasdf:
`disableMultiItemChange` can only apply to `update`, `patch`, `remove`.


eddyystop:
Thanks for the catch!

PR #531, opened 2017-04-18
Added feathers-reduxify-services to frontend adapters.

eddyystop:
feathers-reduxify-services is to Redux like the already listed feathers-vuex is to Vue.

Since its a 3-rd party repo, Feathers isn't responsible for maintaining it.


marshallswain:
@eddyystop this section is for guides & blog posts.  Maybe we should also link to the frameworks section of the Ecosystem page so we don't have to update this in two places.

eddyystop:
Understand.

PR #540, opened 2017-04-22
Fixing old or broken links

teddy-error:
Removed redundant directory path (guides/guides) to fix link. Thankfully, at present, it only doesn't work rather than directing to 404 though I'm surprised Gitbook parsing didn't throw error. I imagine it assumes you're linking externally to some other md file which is hilariously unlikely if you ask me x)


teddy-error:
I ran a scrape of the entire book and this appears to be the only instance of pathing errors resulting in a broken .md link. The other two are valid references to `feathers` Github pages (migrating.md && 1.0 features.md)

eddyystop:
The changes to The Basic guide are OK @daffl.

teddy-error:
Okay. I'm all done for now. 

Worth noting that ba3ea49 needs someone who understands the context of shared authentication better than I to insert a useful link there. From what I could tell, that is either moved into permissions or deprecated- but not sure enough.

teddy-error:
Unclear if oAuth{1,2} modules are still generating changelogs or decidedly removing that feature- it is still in the build file as if it is supposed to occur. So links are not removed entirely in 5486796

daffl:
Thank you for doing this @teddy-error! Also good catch on the changelogs. I added them back and will comment the code back in.

PR #542, opened 2017-04-23
Some small revisions for the CLI/generator guide

daffl:
Some very minor changes for consistency (I'm not entirely sure why the diff is so large actually).


daffl:
I think it's because of the Windows vs. Unix linebreaks.

eddyystop:
You may be right, I wrote some docs on Windows. Makes it hard to check.

What I can see :shipit:.

PR #553, opened 2017-04-27
Improve/guides section

kaiquewdev:
Changed an section to be more expressive and progressive in relation of the level who you can gain across the learning path.


eddyystop:
Thanks.

PR #555, opened 2017-04-27
Improvements/topics

kaiquewdev:
Improvements on communication of each topic because of some fragments presented inside of topics.


ekryski:
First off, thanks for the PR @kaiquewdev! However, as a native english speaker some of the updates you are suggesting don't actually make sense in English.

I'd like to leave them as is for now, but respectfully, thank you for taking the time in putting together this PR. 😄 ❤️

PR #568, opened 2017-05-02
Add HTTPS setup to FAQs

ramblinjan:
This disappeared from the old docs, I think. I just took the text from https://github.com/feathersjs/feathers/pull/33

### DIFF
--- a/faq/readme.md
+++ b/faq/readme.md
@@ -295,6 +295,23 @@ module.exports = {
 }
 ```
 
+## How do I set up HTTPS?
+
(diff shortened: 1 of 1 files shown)


daffl:
It's been right here all along: https://docs.feathersjs.com/api/express.html#https

daffl:
We can keep the FAQ entry but it would be nice to also link to the actual docs.

marshallswain:
Funny.  I did a search for it in my editor before merging.  I'll update it to point there.

PR #571, opened 2017-05-03
Added MostlyJS section and modules

NetOperatorWibby:


marshallswain:
Thanks!

eddyystop:
Maybe this section should be more generatically called `microservices` ?

NetOperatorWibby:
That's what I was going to name it but I saw `DoneJS` so I went with `MostlyJS`. Anything y'all decide is fine with me though!

eddyystop:
I think microservices is more descriptive and other things may end up in
that catgeory. We can look at DoneJS separately.

marshallswain:
yeah.  We're giving client frameworks with more than one entry their own sections.  We tend to generalize on the server packages.

marshallswain:
I've moved it to a Microservices section here: https://github.com/feathersjs/feathers-docs/commit/089a436efedd630a038f4a13c7cd647732113298

Eventually we'll have a guide on how to do this with just Feathers and include realtime.  When that time arrives we can link it from here.

PR #580, opened 2017-05-04
Added warning to avoid an issue with iff, else, iffElse hooks.

eddyystop:
- See https://github.com/feathersjs/feathers-hooks-common/issues/140


ekryski:
I personally agree with this change. I think `this` is pretty esoteric and I'm pretty happy we now have `hook.service`. What do you think @daffl?

eddyystop:
Reopen if needed.

PR #594, opened 2017-05-08
update api/hooks-common/iff|else with array syntax

beeplin:
see https://github.com/feathersjs/feathers-hooks-common/pull/176


eddyystop:
Thanks.

PR #605, opened 2017-05-11
Document env variable escaping

CypherAlmasy:
This is an issue I ran in to recently while setting up authentication.  As far as I can tell, it's not currently clearly documented anywhere that feathers-configuration will not process values that are escaped with backslashes.


ekryski:
Awesome! Thank you @CypherAlmasy! ❤️

PR #606, opened 2017-05-12
Fixed link for feathers-query-filters

Hobbs1210:


eddyystop:
Thanks for the catch and PR!

PR #610, opened 2017-05-14
Fix documentation for setNow/setCreatedAt/setUpdatedAt

hubgit:


eddyystop:
Thanks! I'm always surprised at how much slips by edits.

PR #620, opened 2017-05-21
Documented how to setup authentication inside the chat application

### LINKED ISSUE #601: Comment:Chat-Authentication
Instead of [add the authentication code at the bottom of the file](https://docs.feathersjs.com/guides/chat/authentication.html#getting-a-token), we should update the whole content:

```javascript
const socket = io();
const client = feathers();

// Create the Feathers application with a `socketio` connection
client.configure(feathers.socketio(socket));

// Get the service for our `messages` endpoint
const messages = client.service('messages');

// Configure authentication
client.configure(feathers.authentication({
  storage: window.localStorage
}));

client.authenticate({
  strategy: 'local',
  email: 'feathers@example.com',
  password: 'secret'
}).then((token) => {
  console.log('User is logged in', token);

  // At this point we have a valid token, so we can fetch restricted data.
  messages.find().then(page => page.data.forEach(addMessage));
  messages.on('created', addMessage);
});

// Add a new message to the list
function addMessage(message) {
  const chat = document.querySelector('.chat');

  chat.insertAdjacentHTML('beforeend', `<div class="message flex flex-row">
    <img src="https://placeimg.com/64/64/any" alt="${message.name}" class="avatar">
    <div class="message-wrapper">
      <p class="message-header">
        <span class="username font-600">${message.name}</span>
      </p>
      <p class="message-content font-300">${message.text}</p>
    </div>
  </div>`);

  chat.scrollTop = chat.scrollHeight - chat.clientHeight;
}

document.getElementById('send-message').addEventListener('submit', function(ev) {
  const nameInput = document.querySelector('[name="name"]');
  // This is the message text input field
  const textInput = document.querySelector('[name="text"]');

  // Create a new message and then clear the input field
  client.service('messages').create({
    text: textInput.value,
    name: nameInput.value
  }).then(() => {
    textInput.value = '';
  });
  ev.preventDefault();
});
```

wedneyyuri:
Closes #601

### DIFF
--- a/guides/chat/authentication.md
+++ b/guides/chat/authentication.md
@@ -79,7 +79,7 @@ The returned token can now be used to authenticate the user it was created for b
 
 The Feathers client from the [frontend chapter](./frontend.md) already has authentication (and storing the generated token in LocalStorage) built in and can be used by adding this to `public/app.js`:
 
-```
+```js
 client.configure(feathers.authentication({
   storage: window.localStorage
 }));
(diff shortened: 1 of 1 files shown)


ekryski:
Thanks for the PR @wedneyyuri! This definitely makes it much clearer. Ideally I think we probably want to change the wording to better point out that people should refer to the front-end chapter but this is definitely a great stop gap. Thanks!

daffl:
Hm, I left it intentionally out to show in the framework specific chat guides but it's probably a good call to show the basic idea.

PR #632, opened 2017-05-25
Adds Feathers Vue to ecosystem.

codingfriend1:
Adds https://github.com/codingfriend1/Feathers-Vue to FeathersJS ecosystem.


eddyystop:
Just a thought, are you sure that upper case F won't cause an issue on non Windows?

codingfriend1:
Well, I think it's fine. I built it on a Mac and I had no trouble and it
works on Linux in docker.

PR #637, opened 2017-05-26
Bug fix: postgres example in step 01

zaheraldefai:
Hi featherjs developers,

I like to support you with your wonderful framework. Here is a small bug fix that avoid a "undefined" and a missing dependencies error message.

Best Regards,
Zaher

### DIFF
--- a/examples/step/01/db-connector/3.js
+++ b/examples/step/01/db-connector/3.js
@@ -38,12 +38,16 @@ function databaseConnect() {
     dialect: 'postgres',
     logging: false
   });
-  app.set('sequelize', sequelize);
+  return sequelize;
+  // app isn't defined at that moment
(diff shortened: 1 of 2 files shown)


eddyystop:
Thanks! This might have to wait till the 2nd week of June when daffl
returns.

daffl:
It looks fine to me. Didn't you write that example @eddyystop?

eddyystop:
I didn't add that sequelize code into The Basics guide and I'm rather annoyed by it. I'll have to track down how it got there.

eddyystop:
`examples/step/01/db-connector/2.js` and `/3.js` are not referred to in The Basics guide, and they were never tested. They are there in case we want to expand that part of the guide.

So thanks for testing one of them!

PR #647, opened 2017-05-31
Fix hooks docs

### LINKED ISSUE #645: Comment:Step-Basic-Hooks-1
Comment:Step-Basic-Hooks-1

Hi there, i'm working through the examples section in the documentation and i think the validation hook [here](https://github.com/feathersjs/feathers-docs/blob/master/examples/step/01/hooks/1.js) doesn't work:

When submitting a password that doesn't meet the requirements or omit the password completely, the user still gets created without errors.

Also, when starting the server i get this error `DEPRECATED. Use discard. (remove)` which i wasn't able to track down. I also get `Feathers application started on port 3030` and everything else seems to work fine...

jwahdatehagh:
Fixes the example code referenced in the documentation for hooks (part 1 & part 2).
Fixes #645.

### DIFF
--- a/examples/step/01/common/public/feathers-app-del.js
+++ b/examples/step/01/common/public/feathers-app-del.js
@@ -3,19 +3,19 @@ let jackId;
 const users = feathersClient.service('/users');
 
 Promise.all([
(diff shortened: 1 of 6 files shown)


eddyystop:
Wonderful. Thanks!

PR #654, opened 2017-06-06
Removing obsolete/nonexistent mongoose links

dnszero:
The feathers-mongoose-advanced readme says to use feathers-mongoose.
The feathers-mongoose-service github redirects to feathers-mongoose.

Removing both of these as they clutter the list and make the useful info harder to find.


marshallswain:
Thanks!

PR #657, opened 2017-06-07
Update server.md

cranesandcaff:
Make it clearer that the JWT options need to be passed in on the `jwt` key


marshallswain:
ah, good catch!  Thanks.

PR #659, opened 2017-06-07
Clarified remaining hooks all run once hook.result is set.

eddyystop:


guides/step-by-step/basic-feathers/writing-hooks.md
```diff
@@ -198,9 +198,8 @@ export default function (cache) {
 };
 ```
 
-Feathers checks after every `before` hook if `hook.result` is set.
-If so, Feathers skips any remaining `before` hooks and the database call,
-and starts running the `after` hooks.
+Feathers will not make the database call if `hook.result` is set.
+Any remaining before and after hooks are still run, along with the after hooks..
```
daffl: Shouldn't this be

> Any remaining before and after hooks are still run.

Or

> Any remaining before hooks are still run, along with the after hooks.

(Also with on less `.`)?

eddyystop:
I was thinking hook.result could be changed in the after hooks. Hence "remaining".

eddyystop:
Thanks. Fixed in https://github.com/feathersjs/feathers-docs/pull/661

PR #662, opened 2017-06-07
Update processing.md

paul42:
making 'text' a field with a property on the hook.data object. I'm not certain if previous version was runnable code?


marshallswain:
Thanks for the PR.  The code is actually fine.  It's ES6 shorthand object property syntax.

paul42:
Ah, thanks for letting me know, I'm not familiar with the ES6 too much, so I appreciate the pointers!

PR #707, opened 2017-06-26
Add note about searching documents

arve0:
Ref https://github.com/feathersjs/feathers/issues/334#issuecomment-260834107:

> We will add documentation for searching to the adapters individually.


ekryski:
Thanks @arve0! 🙌

PR #714, opened 2017-06-30
Incorrect variable name

THEtheChad:
Was reading through the example usage with sequelize and noticed that the variable name was incorrectly labelled as `sequelize` when it was obviously meant to be `service`.


marshallswain:
Thanks!

PR #715, opened 2017-06-30
Missing call to instantiate in example

THEtheChad:
Tried to run this example and kept getting an error telling me that `configure` was not a method on `app`. Thought it was related to a change in Express 4.0. Realized that I'm a dummy and that the example code was missing a call to instantiate the app.


marshallswain:
Thanks!

PR #729, opened 2017-07-02
Add feathers-authentication-keystone to ecosystem

alkurbatov:
As it was discussed with @marshallswain :

This pull request adds a link to the new feathers-authentication-keystone plugin which uses Openstack Keystone to authenticate a user in a feathersjs application and a typical howto.


marshallswain:
Thanks @alkurbatov

PR #809, opened 2017-08-14
Indicate that service.emit takes a `hook` argument

jsphpl:


daffl:
Thank you. I think the correct signature here would be `emitter.emit(eventName[, ...args])` which is the same as indicated in the [NodeJS EventEmitter .emit](https://nodejs.org/api/events.html#events_emitter_emit_eventname_args) documentation.

daffl:
Closing since we haven't heard back in a while but always open to a new PR with the suggested changes.

PR #824, opened 2017-09-01
change httpServerConfig to expressServerConfig

nathandial:
Changing the code snippet from httpServerConfig to expressServerConfig to match the example code.


eddyystop:
Thanks for the catch!

PR #841, opened 2017-09-13
Correct the call signature for createQuery

dschnare:
The createQuery function is documented incorrectly as accepting a query object directly, where as it actually only accepts a params object (i.e. object of the shape `{ query: {...} }`.


ckanthony:
thanks for the update!  The old behavior is broken since the upgrade of feathers-knex package.

daffl:
Dangit, I totally missed that this would've been a breaking change and we didn't have tests to cover this.

ckanthony:
nvm bro, thanks for bringing this amazing stuff to the world!

PR #844, opened 2017-09-14
Update readme.md

musicformellons:
link does not work


daffl:
Fixed now. For the future, this would probably make a better issue than PR since it wasn't really anything merge-able.

PR #849, opened 2017-09-17
Add example of modifying token payload

teddy-error:
Saw this answered by @daffl [here](https://stackoverflow.com/questions/45974114/how-can-i-set-the-sub-claim-of-a-jwt-in-feathersjs)

Seems like a common enough scenario to include in the docs. I believe new users might otherwise unnecessarily find this process confusing since there isn't any code explicitly setting the userId claim within the generated `authentication.js`.

### DIFF
--- a/api/authentication/jwt.md
+++ b/api/authentication/jwt.md
@@ -101,6 +101,23 @@ class CustomVerifier extends Verifier {
 app.configure(jwt({ Verifier: CustomVerifier }));
 ```
 
+### Adding Entries to Claim Payload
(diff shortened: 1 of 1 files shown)


teddy-error:
Oh, I looked at #830 and I see that this information is in the recipe section. I will close this pull request unless we'd like to instead add a link to the recipe from the JWT page in lieu of the snippet I added.

daffl:
If https://docs.feathersjs.com/guides/auth/recipe.customize-jwt-payload.html contains all the information we can close this (thank you for the PR though). A new PR linking to the guide would be definitely welcome.

PR #852, opened 2017-09-24
Fix small typo

jakobvogt:
Added a missing quotation mark.


eddyystop:
Thanks for the catch!

PR #853, opened 2017-09-24
Preserve the name in the message

AmazingDesign:
Preserve the name in the message in processMessage hook, if we doesn't  we will get undefined displayed as a sender name in chat app, because the name will be deleted in this hook.


daffl:
Is the `name` needed in the guide?

PR #859, opened 2017-09-28
Hoisting warning on step01

rayfoss:
Especially as it's going out of fashion with let and const

### DIFF
--- a/examples/step/01/db-connector/1.js
+++ b/examples/step/01/db-connector/1.js
@@ -20,7 +20,9 @@ Promise.all([
   users.create({ email: 'jane.doe@gmail.com', password: '11111', role: 'admin' }),
   users.create({ email: 'john.doe@gmail.com', password: '22222', role: 'user' }),
   users.create({ email: 'judy.doe@gmail.com', password: '33333', role: 'user' })
-]) //! [create] /// [results]
+])
+//! [create]
+/// [results]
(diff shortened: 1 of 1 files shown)


eddyystop:
Why would we want to have those `hoisted` comments? 

The `//!` and `///` are used by a gitbook plugin to include code from the file into the document. The merging of 2 of them onto one line will break the docs.

rayfoss:
As I was reading the code those to a second to process... "did I miss something or this not defined"

Promise.all seemed pointless... unless you read 3 lines down to see it continues.

dehamzah:
IMHO, I think you don't need to add comment hoisted here too, since in the documentation it clearly says that [our intended audience](https://docs.feathersjs.com/guides/step-by-step/basic-feathers/database-connector.html#we-require-our-dependencies)

> Readers should have reasonable JavaScript experience, ...

And for this,

> As I was reading the code, those to a second to process... "did I miss something or is this not defined"

I think its not going to be like this if we read the [documentation](https://docs.feathersjs.com/guides/step-by-step/basic-feathers/database-connector.html#we-start-an-instance-of-feathers-and-define-its-services) sequentially like i did for example

daffl:
I agree. Thank you for the PR @rayfoss but I think we'll leave this as it is.

PR #860, opened 2017-09-28
grammar

rayfoss:
I really like soup.


eddyystop:
I'm OK with the offline-first change.

PR #863, opened 2017-10-02
Fix some typo in documentation

dehamzah:
Changelog:
- The example code using `users` object, but the explanation using `user` object
- Add missing Ajv dependencies in example Hooks


eddyystop:
Thank you!

PR #877, opened 2017-10-20
Auth jwt recipe typo on test prop

josx:
There is a typo in test prop beacuse before you customize the jwt with { test: 'test' }


daffl:
Good catch!

PR #880, opened 2017-10-22
Updated `keep` hook docs to warn about it not having any magic provid…

Mattchewone:
…er checks built in

### DIFF
--- a/api/hooks-common.md
+++ b/api/hooks-common.md
@@ -470,11 +470,19 @@ app.service('posts').before({
 });
 ```
 
+> **ProTip:** This hook will always keep the fields,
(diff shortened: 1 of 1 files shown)


Mattchewone:
@eddyystop I just added the warning and a `proTip` example like the `discard` hook. 

Thanks

eddyystop:
Do you think it might better to say this hook will remove any fields not specified even if the service is called on the server. The present wording may sound innocuous to someone reading quickly.

Mattchewone:
@eddyystop Just updated, hopefully that is a little clearer.

eddyystop:
Thanks.

PR #885, opened 2017-10-24
add a recipe for custom auth strategies

subodhpareek18:


subodhpareek18:
Just checked the deploy, the recipe is not appearing on the sidebar actually.

daffl:
The sidebar links are in https://github.com/feathersjs/docs/blob/master/SUMMARY.md

subodhpareek18:
Ready for reviewing 🙏

FreeLineTM:
Will the API key strategy work the same for socket.io?

subodhpareek18:
@FreeLineTM this exact example might not work, as they don't have headers. But as you can see the idea was to be able to create any custom auth module that suits your setup. You can pass a key in the body if you want and return a jwt with that, it's up to you. The two examples shared above were just the proof of concept, of creating JWTs or authenticating in a  standard feathers/passport way.

FreeLineTM:
Alright, thanks! @subodhpareek18

daffl:
Awesome, thank you!

PR #893, opened 2017-10-31
restrictToRoles handles singular string roles

rayfoss:
If you chose to use a single role per user as a string, it will work perfectly. The docs don't mention this. I found out about that it works after starting to fork this.


marshallswain:
Thanks @rayfoss!

PR #896, opened 2017-11-01
feathers-errors and auth-client links broken

rayfoss:
Last of the bunch.

Only exception is https://docs.feathersjs.com/api/client.html#universal-isomorphic-api
which links to https://github.com/feathersjs/feathers/blob/master/src/client/express.js

I can't see where it moved based on content.

> ## Universal (Isomorphic) API
> 
> The Feathers Client uses the same [Application API](./application.md) as is available on the server.  It is extremely lightweight, however, with Express having been replaced by a [thin wrapper](https://github.com/feathersjs/feathers/blob/master/src/client/express.js).

### DIFF
--- a/api/authentication/client.md
+++ b/api/authentication/client.md
@@ -59,7 +59,7 @@ To enable `localStorage` on the client, be sure to set `storage: window.localSto
 
 After configuring this plugin, the Feathers client will have a few additional methods:
 
-### `feathersClient.authenticate(options)` [source](https://github.com/feathersjs/feathers-authentication-client/blob/master/src/passport.js#L136)
+### `feathersClient.authenticate(options)` [source](https://github.com/feathersjs/feathers-authentication-client/blob/master/lib/passport.js#L136)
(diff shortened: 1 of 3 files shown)


daffl:
Thank you. I'm wondering how those source links ended up there, I'm usually not a fan of putting them in for exactly this reason 😉 
I think I'm going to suggest to actually remove them because in the future file names or the code might change and then we have the whole problem over again.

rayfoss:
@daffl come to think of it, inline drop down examples would be nice. It would also lower the barrier to adding them. For instance, symfony often shows 4 different ways of doing the same thing: http://symfony.com/doc/current/templating.html#linking-to-pages

it would be great to have a lower barrier to entry for examples, and thus get examples from more people. And also welcome links to gist for the particularly long ones.

daffl:
Are we talking about links to the actual source code or example code?

rayfoss:
Sample code. For instance, the validation doc is really anemic. I would love it if it included links to partial and/or complete examples.

It doesn't seem under the current guidelines that for instance, the developer of the

PR #900, opened 2017-11-02
Fix 'iff' conditional hook example...

pkreipke:
`... !isProvider('service')...` on line 484 should use `'server'` not `'service'`

Reported in https://github.com/feathersjs/docs/issues/890


eddyystop:
Thanks!

Issue #564, opened 2017-04-30
Search is broken on docs site


petermikitsh:
<img width="1072" alt="screen shot 2017-04-30 at 3 52 32 pm" src="https://cloud.githubusercontent.com/assets/1571918/25568683/17c6b2dc-2dbd-11e7-9dc4-2dd5d82cf31c.png">

I'm pretty sure something should come up for the query `auth`. Regardless, based off the output, it seems to be ignoring characters after the first.

daffl:
Confirmed. This is the error being thrown:

<img width="855" alt="screen shot 2017-04-30 at 9 19 53 pm" src="https://cloud.githubusercontent.com/assets/338316/25571968/ced058c0-2dea-11e7-91f1-bbfddd772c26.png">

teddy-error:
Sometimes it just doesn't work at all. It also doesn't clear the blue when you exit the search box or even navigate to a new page.

marshallswain:
This is a GitBook issue.  I couldn't figure out which version of gitbook we're using.  Similar issue: https://github.com/GitbookIO/gitbook/issues/497

daffl:
Maybe switching to something like https://www.npmjs.com/package/gitbook-plugin-algolia might be the better way to solve this.

daffl:
My suspicion is that this is because of emoticon in the text but I am not sure. I have disabled search for now.

petermikitsh:
I actually just used search this morning -- it worked after the first character. You need to type in two characters or more. Was on latest chrome, not sure if issues remain for other browsers.

daffl:
@petermikitsh Did you see the error?

petermikitsh:
Didn't have the console open :(

The UI looked similar to what @teddy-error's screenshot shows after the first letter is typed -- second+ characters began to populate search results.

teddy-error:
[Other FOSS projects appear to be using the search plugin without issue](https://github.com/Vue-cd/vue_all/blob/b588d0082f3a4997c6dcc66a276fe8610274a6d6/vuex%2B/docs/book.json). This may suggest that our documentation has some unique trait(s) which are causing the issues.

daffl:
Probably. As I mentioned, so far at least the page at https://docs.feathersjs.com/guides/about/vs.html is causing this issue.

teddy-error:
I was able to resolve the issue and will create a pull request shortly.

teddy-error:
[...]

PR #907, opened 2017-11-06
Replace literal emojis with placeholders

teddy-error:
Markdown files containing literal/unicode emojis cause issues with `lunr` indexing. Attempting to omit these files from indexing as described in the [lunr documentation](https://github.com/GitbookIO/plugin-lunr) proves ineffective and does not prevent the exception described in #564.

While simply removing the emojis would presumably be a solution we can also simply replace them at compile time using [a plugin](https://github.com/codeclou/gitbook-plugin-advanced-emoji)

This pull request replaces all emojis within the Feathers documentation with their equivalent placeholders. I.e.: `:+1`. Because indexing occurs at compile time they will be parsed in this state and then transformed into literals by the plugin.


teddy-error:
You can see the resulting site after changes at https://teddy-error.gitbooks.io/feathersjs

Additionally, I should note that the 🤓 emoji representing **Community Maintainer** in `vs.md` is not available within emojify and is instead represented by 💞 after these changes.

teddy-error:
Sure. Tell @ekryski owes me a beer for fixing all these emojis he added though 😏

daffl:
He's actually just brewing some 🍻  😄 

Thanks again for doing this.

PR #913, opened 2017-11-09
Buzzard api docs: rename `hook` object to `context` object; a few typo fixes

beeplin:
I went through the whole docs repo and tried my best to rename all `hook`s as the object name to `context`  while keeping `hook`s as function name untouched. You guys might need to take some time to review this PR to see if there are any mistakes. :)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "editor.formatOnSave": false,
+  "eslint.autoFixOnSave": true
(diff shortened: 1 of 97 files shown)


eddyystop:
Wow.

beeplin:
In fact it just took less than 2 hours with the help of some regex search/replace tools.

daffl:
Thank you for doing this!

PR #922, opened 2017-11-17
Avoid long timeouts in React Native when using feathers-socketio

vgeorge:
The default values of `pingInterval` and `pingTimeout` in Socket.io are longer than one minute and will fire warnings in React Native when using feathers-client with Socket.io. This commit adds instructions to the client docs on how to avoid these warnings properly.

### DIFF
--- a/api/client.md
+++ b/api/client.md
@@ -112,6 +112,18 @@ messageService.create({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


daffl:
Interesting. Definitely worth adding. Would you mind adding this to http://buzzard.docs.feathersjs.com/api/client.html#react-native (also with the "Edit this page" link at the top)? That way I don't have to merge it back.

vgeorge:
Done: https://github.com/feathersjs/docs/pull/948

daffl:
Closed via https://github.com/feathersjs/docs/pull/948

PR #933, opened 2017-12-05
Update philosophy.md

jraut:
User abhisekp was wondering in Slack: "I thought `callbacks` were deprecated. Why does it show it in docs and not specify that they're deprecated?"


daffl:
Thank you.

PR #937, opened 2017-12-06
Update hooks.md with cancel hooks note.

aplb:
This question arises from time to time so I decided to add this section that mentions it.

### DIFF
--- a/api/hooks.md
+++ b/api/hooks.md
@@ -326,3 +326,27 @@ app.hooks({
   }
 });
 ```
+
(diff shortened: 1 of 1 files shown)


daffl:
Thank you! I think it's a good call but would suggest that it makes more sense as an example in the [iff hook documentation](https://feathers-plus.github.io/v1/feathers-hooks-common/#iff).

eddyystop:
I've made a note to add it to those docs.

PR #946, opened 2017-12-14
Fix links

bertho-zero:


daffl:
Thank you. I have a feeling there may be more 😄

PR #948, opened 2017-12-15
Avoid long timeouts in React Native

vgeorge:
The default values of pingInterval and pingTimeout in Socket.io are longer than one minute and will fire warnings in React Native when using feathers-client with Socket.io. This commit adds instructions to the client docs on how to avoid these warnings properly.


daffl:
Great, thank you!

PR #953, opened 2017-12-18
Add extension to executable JS.

HoverBaum:
It might seems confusing to some (like myself just now) to have the server file here without the `.js` extension. While it runs fine I feel the guide will benefit from being easy to understand and keeping to a convention of explicitly stating file endings in node calls.


eddyystop:
Thanks!

PR #956, opened 2017-12-19
Fix typos: "type" --> "typ"

morphatic:
Change examples to match [the JWT spec](https://tools.ietf.org/html/rfc7519#section-5.1)


daffl:
Great, thank you!

PR #957, opened 2017-12-19
Fix hook-flow.jpg: after all hook runs before other after hooks

beeplin:


eddyystop:
The after-all hooks should run when the after hooks for the individual services are completed.

Have you determined this with with Auk or Buzzard? Could you check it again?

beeplin:
@eddyystop This graph is for the following hook configuration:
image

I am not sure if I understand this right, but according to the docs, for the hooks on a single service (not the application-wide hooks), after-all hooks should run before other after hooks:

```
Note: all is a special keyword which means those hooks will run before the method specific hooks in this chain.
```

beeplin:
I recalls that we have discussed about this issue before: https://github.com/feathersjs/feathers/issues/533

daffl:
This is correct, thank you very much for doing this! I started looking at the Sketch file to fix this but I'm not friends with software like that 😉

PR #959, opened 2017-12-19
Fix typo

adrianolsk:
No -> Now


daffl:
Thank you! Those are brand new so there might be one or two in there 😉

PR #972, opened 2017-12-21
Explicitly mention Feathers error module install

aral:
The code sample under hooks will not compile without the reader first installing the Feathers error module. This makes that explicit and avoids an error on first run.


daffl:
Totally makes sense, thank you!

PR #980, opened 2017-12-22
Fix two typos

aral:


daffl:
Thank you for catching all of those! I really thought I enabled spellcheck in my editor but it seems I did not (I make the `connection` typo all the time ;)

PR #1003, opened 2018-01-01
sudo required when using -g

edwardsmarkf:


daffl:
`sudo` should not be required, it might be a misconfiguration of your Node installation. I recommend [nvm](https://github.com/creationix/nvm) which allows to quickly install new versions and runs everything locally to your user.

PR #1005, opened 2018-01-02
fix channels example comment

uinz:
```js
// user.rooms.forEach(room => app.channel(`rooms/${room.id}`).join(channel))
```
=>

```js
// user.rooms.forEach(room => app.channel(`rooms/${room.id}`).join(connection))
```


daffl:
Thank you!

PR #1010, opened 2018-01-03
eliminate 'localhost' from website scripts

edwardsmarkf:
when running over a network, the value of "localhost" has to be constantly changed, which is way too easy to forget for a newbie.

same change would be in in https://docs.feathersjs.com/guides/basics/clients.html

### DIFF
--- a/guides/basics/real-time.md
+++ b/guides/basics/real-time.md
@@ -152,7 +152,7 @@ Then we can initialize and use the socket directly making some calls and listeni
 /* global io */
 
 // Create a websocket connecting to our Feathers server
-const socket = io('http://localhost:3030');
+const socket = io(':3030') ;  //  or use io('http://localhost:3030');
 
 // Listen to new messages being created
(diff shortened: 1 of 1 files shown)


daffl:
Thank you. In that case, wouldn't your change not work just the same as the original though?

edwardsmarkf:
in my quest to learn feathers, i am using google-console vps.  none of the "localhost" commands used in a webpage scripts work for me until i replace "localhost" with the server IP.  

it seems to me by eliminating the "http://localhost" part, your webpage scripts will work in all situations.

daffl:
Got it. I think it's a good suggestion but I'm wondering if only having the partial URL in there could also be confusing (`:3030` is a valid URL but it does look kind of strange and might throw some people off).

daffl:
I think we'll keep it at localhost since the guides are expected to be run locally. If you are running on a different domain or port it has to be adjusted anyway. Thank you for the PR though!

PR #1027, opened 2018-01-11
Fix typo & some wording

SassNinja:
I'm just getting into feathers and thus went through the basics guide (nice docs btw).

In the course of this I've find a typo (`contians` instead of `contains`) and some wording that could be improved (missing single quote & `a the` link)

This PR fixes it.


daffl:
Perfect, thank you!

PR #1034, opened 2018-01-16
added client authentication events

flyboarder:


daffl:
Thank you. I wanted to fix them to align with normal auth but let's get those in so that they are at least documented.

flyboarder:
@daffl I agree, these names are confusing it would be nice if the client had things like `login/logout`

PR #1037, opened 2018-01-17
Fix misspelling of @feathersjs/authentication-jwt

bnielsen1965:


daffl:
That's what I get for global find and replace 😉  Thank you!

PR #1039, opened 2018-01-19
Update configuration.md to accurately reflect configuration behavior

otang:
"When the NODE_ENV is not development" is not accurate. Even if NODE_ENV==='development' the `config/development.js`configuration will be loaded.


daffl:
Good catch, thank you!

PR #1043, opened 2018-01-24
Service as a class must be instantiated with the "new" keyword

osenvosem:


daffl:
That is true, thank you!

PR #1047, opened 2018-01-28
Fix typo of installing command into RN project.

PoshHsu:
The namespace should be @feathersjs instead of @feathers.


daffl:
Thank you! I kept making that mistake all the time 😉

PR #1058, opened 2018-02-05
Fix examples

bertho-zero:


daffl:
As always, very much appreciated. No idea how I missed those, I think it was some last minute copy & paste. Thank you!

PR #1061, opened 2018-02-07
Add Telegram chat

vodniciarv:
Add Telegram chat


daffl:
Great, thanks for adding those!

PR #1073, opened 2018-02-13
Improve readability

claforte:


daffl:
If you would like to review more than one page, I'd recommend to clone down this repository and submit all the changes at once (you can preview them by running `npm install` first and then `npm run build` which will generate the docs into the `_book` folder). That makes it easier to review and merge.

claforte:
Thanks David. If you don't mind, I will continue to update pages one-by-one as needed, as I don't expect to review a lot more in the short term. Thanks,

Christian

PR #1083, opened 2018-02-21
Changed ++this.currentId to this.currentId++

sorja:
This is to be more concise with the rest of the material, later on we have `id: 0`, but having ++ before the currentId increments the variable before using it.


daffl:
Does the other way cause any issues with the subsequent documentation? I'd prefer having the id start at 1 because checks like `if(id)` with an id of `0` will fail.

sorja:
Yes, the documentation mentioned id: 0, I changed it to id: 1 now.

daffl:
Ah, good catch, thank you!

PR #1087, opened 2018-02-23
Update services.md

bobinush:
Since line 105 is:
id: ++this.currentId
the value is increased before assigned


daffl:
Thank you! I guess we didn't catch them all in #1083

PR #1094, opened 2018-03-09
Fixed the parameters shown in example.

gauravdhiman:
As per the definition of `joinChannels`, parameters should be `user` and `connection`. Corrected the same in docs.


daffl:
Thank you!

PR #1098, opened 2018-03-14
Demonstrate how to change the config file location

aral:
I had to spend a bit of time hunting this down from a dependency’s documentation. Adding it here so others won’t have to.

If there is a more direct way to achieve the same result with the @feathersjs/configuration API directly, please document that instead. Looking at the code, I couldn’t find it.


daffl:
I've been wondering if it defeats the purpose of environment variables when setting them in code but I think in this case it makes sense since the location will probably stay the same in different environments. Thank you!

PR #1099, opened 2018-03-14
Add failure redirect to authenticate example

Jaden-Giordano:
Adding a secondary example in case the developer wants to redirect the user on a failed authentication to a protected route.


daffl:
Thank you! Related: https://github.com/feathersjs/authentication/issues/648

PR #1101, opened 2018-03-16
fix database adapters broken link

thakkaryash94:
it is returning 404 for readme.md page. I think right doc path is adapters.md.


daffl:
Thank you!

PR #1103, opened 2018-03-17
Update real-time.md

YuriGor:
Serving static files added.

### DIFF
--- a/guides/basics/real-time.md
+++ b/guides/basics/real-time.md
@@ -109,6 +109,9 @@ app.use('messages', memory({
   }
 }));
 
+// serve static files to access them in browser
+app.use(express.static('public'))
+
 // Set up an error handler that gives us nicer errors
 app.use(express.errorHandler());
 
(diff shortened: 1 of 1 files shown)


daffl:
Hm, maybe it should be changed at this step or even before. The original idea was that the web server that is being [set up when getting started](https://docs.feathersjs.com/guides/basics/starting.html#in-the-browser) is still running.

YuriGor:
I think many people, like me, have no enough free time to go through this tutor at once, so it's better to make each article self-sufficient. If it's ok to use same express instead of http-server, then we will need no CORS headers later in https://docs.feathersjs.com/guides/basics/clients.html
Well I can change other articles too and submit PR

daffl:
I was going back and forth between those two approaches but I think you are right, it makes sense to have each page work standalone. Maybe you can double check but I actually think this one is the case where it really matters right?

YuriGor:
Done. 
Check it please, English isn't my native.

daffl:
Thanks for doing this @YuriGor. Not having to deal with the CORS stuff is actually really nice. I'll pull this in and make some additional small changes (e.g. the `starting.html` chapter should work both, in the browser and Node which won't be possible when using it with Express) but then we'll :shipit:

PR #1106, opened 2018-03-18
typo

musicformellons:
successfull does not exist, see https://english.stackexchange.com/questions/4617/successfull-successful-is-this-a-uk-us-difference


daffl:
Ha, I always make that mistake 😉 Thank you!

PR #1110, opened 2018-03-19
Add access to `services` array

bitflower:
As pointed out by @eddyystop on Slack ;-)

### DIFF
--- a/api/client.md
+++ b/api/client.md
@@ -49,6 +49,10 @@ messageService.on('created', message => console.log('Created a message', message
 messageService.create({
   text: 'Message from client'
 });
+
+// Loop though all services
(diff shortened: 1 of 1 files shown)


Palivonas:
Should be more descriptive than "loop though all services" - it actually fetches all the data from all the registered services. Maybe just make it log a list of registered services into console instead?
`.forEach` should be used in this case, not `.map`

daffl:
Thanks for the PR @bitflower! I should probably add `app.services` in the [Application API](https://docs.feathersjs.com/api/application.html) instead though. It is also an object with `path: serviceObject` so the correct way to loop would be:

```js
Object.keys(app.services).forEach(path => {
  const service = app.service(path);
});
```

PR #1114, opened 2018-03-25
Update recipe.mixed-auth.md

MarcGodard:
context needs to return


daffl:
It is optional but definitely the recommended way to do it. Thanks!

PR #1117, opened 2018-03-28
Fix code snippets being shown incorrectly

bernardobelchior:


daffl:
Thank you but the Codetabs plugin is being used by GitBook to generate tabs in the actual page (https://docs.feathersjs.com/api/client/rest.html) and looks like this:

screen shot 2018-03-28 at 12 29 35 pm

The pages in this repository are in general valid Markdown but ultimately intended to generate the documentation at https://docs.feathersjs.com not to be read in the repository directly.

bernardobelchior:
Ok, thank you for the clarification!

PR #1118, opened 2018-03-30
Update docs about custom handler

chrisronline:
This is necessary when the client and server reside on different domains.


daffl:
Thank you!

PR #1119, opened 2018-04-03
Swap the order of the hook registry example and diagram

akagomez:
This may have just been me, but I struggled to grok the "Registering Hooks" flow diagram. It wasn't until I gave up and scrolled passed that it became clear from context below. 

My recommendation here is to invert the order of the code/diagram and revise the wording.


daffl:
Totally makes sense, thank you!

PR #1124, opened 2018-04-09
Fix broken inline hook execution diagram.

nicksergeant:
The hook execution diagram is appearing as inline code in the example code above it:

This change fixes that and embeds the diagram properly.


daffl:
Thanks! Looks like it snuck in there with #1119

PR #1125, opened 2018-04-09
Clarify databases chapter usage of NeDB.

nicksergeant:
It seems that the [databases chapter](https://docs.feathersjs.com/guides/basics/databases.html) doesn't actually demonstrate usage of NeDB with `feathers-nedb`. Unsure if this used to be detailed or if it hasn't been written up yet, but this PR might help remove some confusion.


daffl:
Thank you. I think I was planning to use NeDB but then decided against it.

PR #1129, opened 2018-04-14
Docs to allow multiple express mw functions to be passed into `app.use`

eXigentCoder:
This documentation update is for the pull request in [feathersjs/express](https://github.com/feathersjs/express/pull/25). It shows the added support for passing in a middleware array to `app.use` and includes a code example of how to use it.


daffl:
Thank you for also getting the docs updated!

PR #1131, opened 2018-04-25
Add missing default options

robbiedhickey:
In a company project we recently had a bug that came down to the entityUsernameField's default value. I noticed that in the deprecated feathers-authentication-local package these options are documented. Also confirmed that the authentication-local plugin still checks for the presence of entity*Field options.


daffl:
You're right, those definitely should've been deprecated. Makes sense keeping it in the docs until it is removed.

PR #1142, opened 2018-05-11
Update create-react-app doc

eric-burel:
Fixed a missing link and made the doc more obvious


daffl:
Thanks. Just made a minor edit. If you import `@feathersjs/client` the modules will be available as `feathers.rest` already. `@feathersjs/client/rest` only should be imported if that is the only thing you are using.

eric-burel:
Thanks it makes thing clearer indeed

PR #1145, opened 2018-05-18
Fixed typo in "Guides > The Basics > Hooks #What's next"

temowemo:
[Link to section](https://docs.feathersjs.com/guides/basics/hooks.html#whats-next)

What do you think?

From:
`...can be used middleware for service method calls...`
To:
`...can be used as middleware for service method calls...`


daffl:
Thank you!

PR #1148, opened 2018-05-21
Grammar suggestion

temowemo:
I think the following:

> Channels determine which real-time events should be sent to which client. For example, we want to only send messages to authenticated users or users in the same room. For this example we will just enable real-time functionality for all connections however:

Can be changed to:

> Channels determine which real-time events should be sent to which client. For example, we might want to only send messages to authenticated users or users in the same room. For this example however, we will just enable real-time functionality for all connections:

Thoughts?

### DIFF
--- a/guides/basics/real-time.md
+++ b/guides/basics/real-time.md
@@ -60,7 +60,7 @@ app.listen(3030);
 
 ## Channels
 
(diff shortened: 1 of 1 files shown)


daffl:
👍

PR #1150, opened 2018-05-21
More precision about the auth setup/transport setup order

eric-burel:
Make the config order more obvious.
I fell in this trap and lost an hour or two so I guess it deserves to be emphasized ^^ Sorry if the phrasing/terms are not quite precise though.


eric-burel:
Nice, thanks for your reactivity!

PR #1159, opened 2018-06-07
add Error Handler recomendation

toddgeist:
The intent is to suggest a way to handle errors so that sensitive information doesn't flow back to the client.  I am not sure my example is the best method. But I think something like this should be included in the docs.


daffl:
Thank you. Adding that section makes sense, I'd probably add some details with links to the [Express error handler](https://docs.feathersjs.com/api/express.html#expresserrorhandler) and mention error hooks.

toddgeist:
ok added some details with links :-)

PR #1168, opened 2018-06-21
use explicit key

winchell:


daffl:
Tank you for the pull request @winchell! All documentation pages use standard ES6 syntax [as mentioned in the prerequisites](https://docs.feathersjs.com/guides/basics/setup.html) so I would prefer not to change it in this one case though.

PR #1169, opened 2018-06-22
Remove redundant const

winchell:


daffl:
Makes sense, thank you!

PR #1175, opened 2018-07-10
Update real-time.md

greenspray9:
In https://docs.feathersjs.com/guides/basics/rest.html it was mentioned that - 
"Important: In Express an error handler, here app.use(express.errorHandler());, always has to be the last line before starting the server."

but the current documentation calls socketio after errorHandler. Or is it not important for this case when the errorHandler is called.


daffl:
It's not important in this case since the Socketio does not add any of its own middleware but it is indeed a good change, thank you!

PR #1176, opened 2018-07-10
Clarify after hooks are still run when SKIP is returned in a before hook

eddyystop:
The present text is ambiguous.


daffl:
Well, I learned something new then 😉 Thanks!

PR #1178, opened 2018-07-15
Fix reference to env-specific config file

barraponto:
It mentions the file as ` .js` but it's actually a json file.
(otoh, it'd be amazing to have a js configuration example in this page).


daffl:
Makes sense, thanks!

PR #1183, opened 2018-07-23
Change function to middleware in options docs

barraponto:
In the documentation renderer, `function` breaks syntax highlighting.
Middleware is a common term for Express developers, though we might want to hint at the signature `(req, res, next) => {}` or `(err, req, res, next) => {}`.
It's an improvement over `function` anyway.


barraponto:
Here's a link to the live docs to see the syntax highlighting issues: https://docs.feathersjs.com/api/authentication/oauth2.html#options

PR #1185, opened 2018-07-27
Invalid example of "app.hooks" error

danawoodman:
Should instead be passed a `HookObject` which includes an `all` key. The example as-is broken.

### DIFF
--- a/guides/basics/hooks.md
+++ b/guides/basics/hooks.md
@@ -200,8 +200,10 @@ A good use for application hooks is to log any service method call error. The fo
 
 ```js
 app.hooks({
-  error: async context => {
-    console.error(`Error in '${context.path}' service method '${context.method}'`, context.error.stack);
+  error: {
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request @danawoodman! If you have a [look here](https://docs.feathersjs.com/api/hooks.html#registering-hooks) you can see that it is also a valid way to register hooks though.

danawoodman:
According to the type definitions for Typescript that isn’t valid. Maybe the type definitions are wrong?

daffl:
Yes, that is an oversight in the typings then.

danawoodman:
Anything actionable here?

daffl:
Yes, you could track it as an issue in https://github.com/feathersjs-ecosystem/feathers-typescript or even better, update the definitions at https://github.com/DefinitelyTyped/DefinitelyTyped.

PR #1192, opened 2018-08-16
Fix typo

soulofmischief:
The guide references `feathers generate authentication` a second time where I'm pretty sure it means `feathers generate service`.


daffl:
Well, that makes sense. Thank you!

PR #1195, opened 2018-08-27
Possible upgrade to express/socket issue

edwardsmarkf:
i believe we need to add a new line to get this example working, as per   

https://expressjs.com/en/starter/static-files.html
https://www.tutorialspoint.com/expressjs/expressjs_static_files.htm

maybe caused by a newer version of socket.io or express?

### DIFF
--- a/guides/basics/databases.md
+++ b/guides/basics/databases.md
@@ -160,6 +160,8 @@ app.use(express.urlencoded({ extended: true }));
 app.configure(express.rest());
 // Set up an error handler that gives us nicer errors
 app.use(express.errorHandler());
+// Establish the ./public/ directory as a source for static files
+app.use(express.static('public'));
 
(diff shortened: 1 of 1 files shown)


daffl:
We can add this but the example itself does not require static files to be served and I believe the full examples in the Readme of each database adapter do not use it either.

edwardsmarkf:
Thank you - I found this line to be required.  Otherwise, i kept getting the error message **Cannot serve '/'** when running the examples while using the **./public/index.html** and the **./public/client.js** mentioned in the "_In The Browser_" section preceding this example. 
 
Also, without the new line, I believe that extra code is needed in **./public/client.js** to specify where the feathers server is running, since the line of code **const app = feathers();** will not work, unless the port number is specified -- was it something like **const app=feathers(":3030");** ??

PR #1196, opened 2018-08-31
Update authentication.md

batagliao:
Change in section 'Securing the messages service'
running the sample as was, received the error 'authenticate is not a function'
So, I saw that the function is instead in the hooks object. Replaced the call to include the object containing the authenticate methods

### DIFF
--- a/guides/chat/authentication.md
+++ b/guides/chat/authentication.md
@@ -74,11 +74,11 @@ The returned token can then be used to authenticate this specific user, by addin
 Let's restrict our messages service to authenticated users. If we had run `feathers generate authentication` *before* generating other services, `feathers generate service` would have asked if the service should be restricted to authenticated users. However, since we created the messages service first, we now have to update `src/services/messages/messages.hooks.js` manually to look like this:
 
 ```js
-const { authenticate } = require('@feathersjs/authentication').hooks;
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request @batagliao but the existing example is correct. It is using [ES6 destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to pull in the `authenticate` property from `require('@feathersjs/authentication').hooks`.

batagliao:
@daffl Sorry for not notice that. I'm just beginning in Node and it wasn't woking for me. Just wanted to help. Sorry for your time. And thanks for linking in the comment so I can read and learn that topic

daffl:
Totally appreciate the contribution. Basically there is two ways you can write this. The original example which is

```js
const { authenticate } = require('@feathersjs/authentication').hooks;

module.exports = {
  before: {
    all: [ authenticate('jwt') ],
```

Or

```js
const { hooks } = require('@feathersjs/authentication');

module.exports = {
  before: {
    all: [ hooks.authenticate('jwt') ],
```

Both will work the same.

batagliao:
Thanks @daffl

PR #1215, opened 2018-10-23
Update app.js example using messages service

jamesobd:
The feathers-memory library is no longer needed in this app.js example because the library is now used in the messages service.


daffl:
True indeed, thank you!

PR #1216, opened 2018-10-24
Added 'secret' property for authentication options

reggaedit:
Property that seems to be required but is not mentioned in the docs in the authentication options

From [jsonwebtokens](https://github.com/auth0/node-jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback)
`secretOrPrivateKey` is a string, buffer, or object containing either the secret for HMAC algorithms or the PEM encoded private key for RSA and ECDSA. In case of a private key with passphrase an object { key, passphrase } can be used (based on crypto documentation), in this case be sure you pass the algorithm option.


daffl:
Thanks. Super strange why the most important option was missing.

PR #1231, opened 2018-11-25
Fix documentation error about local.hooks.protect

Nico264:
The password field didn't get protected in the documentation example.

The local.hooks.protect hook must be an global after hook and not an after.create one.
The [feathersjs-authentication README](https://github.com/feathersjs/authentication/blob/master/README.md) is already correct but this documentation was not.


daffl:
Ah, yes that makes sense. Thank you!

PR #1263, opened 2019-02-11
introduce $eq and $and

edwardsmarkf:
since both **$eq** and **$and** are part of Sequelize, should they both be referenced here?

i am not sure if they are part of NoSQL, but i should think they are also included with NoSQL.

### DIFF
--- a/api/databases/querying.md
+++ b/api/databases/querying.md
@@ -16,6 +16,14 @@ app.service('messages').find({
     roomId: 2
   }
 });
+
+// Find all unread messages in room #2 using verbose $eq
+app.service('messages').find({
+  query: {
+    read: { $eq: false},
+    roomId: { $eq: 2 }
+  }
+});
 ```
 
 ```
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request but they are not part of the common query syntax which is supported by all database adapters. The common query syntax handles `$eq` and `$and` implicitly.

edwardsmarkf:
oh well....

daffl:
It's been brought up a couple of times that `$and` might be a good addition in the future (not sure about `$eq` since `{ name: { $eq: 'David' } }` is the exact same as querying with `{ name: 'David' }`.

edwardsmarkf:
yes it is exactly the same, but often when learning, when one only learn
the "shortcuts" people miss the big picture.  and when $and +  $ne are used
together for like this:

query {
   $and { name:  { $eq: 'David' },
              city:      { $ne: 'Calgary' },
            }

i think it looks a bit more "balanced" and a bit easier to read.

an old boss of mine used to say "you write it only once, and read it a
thousand times".  he just passed away recently, but his words stick with
me.  that is why i use for example "yum --assumeyes" rather than just "yum
-y" in scripting.

Thank you,

Mark Edwards

PR #1266, opened 2019-02-13
include certification chain file

edwardsmarkf:
i believe the certification chain is now required.  when i tried to omit it, i received an error.  also reference the cert files in ./config/default.json.

### DIFF
--- a/api/express.md
+++ b/api/express.md
@@ -118,16 +118,22 @@ HTTPS requires creating a separate server in which case `app.setup(server)` also
 ```js
 const fs = require('fs');
 const https  = require('https');
+const port = app.get('port');
 
 const feathers = require('@feathersjs/feathers');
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request. I don't believe this has changed or is necessary since [the HTTPS tests](https://github.com/feathersjs/feathers/blob/master/packages/express/test/index.test.js#L204) still appear to be passing.

edwardsmarkf:
hey david --  maybe its only required for https?
https://www.ssllabs.com/ssltest/    i am guessing it will be required soon
enough though.  also, i think its a good idea to keep the cert values in
the config file, imho.

Thank you,

Mark Edwards

daffl:
I'm going to close this one since the previous one has been tested and still works. Thanks for the PR though!

PR #1281, opened 2019-03-23
Altered misleading wording

james2mid:
Sorry if this seems picky or annoying but I believe your wording to be misleading.

Your wording implied that an alternative to the `feathers-blob` module was being covered in this document which is not true. I hope this is clearer.


daffl:
Makes sense, thank you!

PR #1282, opened 2019-03-25
Code example under ## Service Methods malformed

coletain:


daffl:
Thank you for the pull request but this is currently still intended. Is it a plugin to show tabs on the rendered documentation at https://docs.feathersjs.com/guides/basics/services.html

PR #1284, opened 2019-04-01
Crow link check

DesignByOnyx:
This creates a script to validate links within the docs. The script will validate that referenced files exist and that referenced `#anchors` exist. This currently only checks cross-references within the docs and does not check external urls (`^https?://`). External link validation can be added if we want.

The first commit contains the actual script for validating links. The second commit contains all of the fixes reported by this tool.

This script uses a cool set of tools under the [unified](https://unified.js.org/) umbrella:

- [unified-engine](https://github.com/unifiedjs/unified-engine) - the core file processing engine
- [remark](https://github.com/remarkjs/remark) - markdown processor for unified
- [remark-validate-links](https://github.com/remarkjs/remark-validate-links) - link validator plugin for remark
- [vfile-reporter](https://github.com/vfile/vfile-reporter) - reporter for printing errors/warnings to console


daffl:
This is great, thank you! Will be very helpful when doing a full review of the docs before the release.

DesignByOnyx:
Thanks for merging. One note of particular is that this script currently considers redirected links as valid. For example, an older version of the docs contained a root-level "authentication" section (as well as many other root-level sections). These are now being redirected like so:

```
/authentication/*   /api/authentication/server.html
/clients/*          /api/client.html
/configuration/*    /api/configuration.html
# ... many more like this ^^^
```

As of this comment, this script considers any link which references an old file as valid. This is fine for today, but as time moves on we may eventually have root-level sections which we no longer want to be redirected. We might also want the docs to be "pure" and only use up-to-date references. In such a case, simply comment out the `validateRedirects` portion of the script.

PR #1291, opened 2019-05-02
Crow: More authentication documentation (still WIP)

daffl:


api/authentication/service.md
```diff
@@ -0,0 +1,148 @@
...
+
+## configuration
+
+`service.configuration` returns a copy of current value of `app.get(configKey)` (default: `app.get('authentication')`). It is usually initialized through an `authentication` property in a [configuration file](../configuration.md) (default: `config/default.json`) which must at least contain a `secret` property. The default configuration is:
+
+```js
+{
+  entity: 'user', // The name of the entity. Can be `null` when not used
+  service: 'users', // The service to retrieve the entity from
+  jwtOptions: {
```
KidkArolis: This was changed to accessTokenOptions, or not?

KidkArolis:
Would be great to document where custom oauth params go, e.g. I've used in the previous version, scope is probably already correct in grant, but prompt is something that is probably not there by default. This is a custom query param sent to google. I can PR when I figure this out if you're not sure.

```
  prompt: 'select_account',
  scope: 'email openid profile',
```

KidkArolis:
Also, kind of related to docs, one of the few remaining things left for me to port from old auth to 4.0.0 is redirects. E.g. in the old auth, I was using `handler: OAuth2Handler()` to do a custom redirect to a signup flow step 2 (instead of immediately logging in the user):

```
//  based on https://github.com/feathersjs/authentication-oauth2/blob/9a3d27c8a0fa0341e79131670d2ea6c8301d5a3d/lib/express/handler.js
function OAuth2Handler() {
  return async function handler(req, res, next) {
    try {
      const payload = req.feathers.payload
      if (payload.verificationId && payload.verificationToken) {
        // sign up flow!
        const { verificationId: id, verificationToken: token } = payload
        redirect(res, `${config.google.signupRedirect}?id=${id}&token=${token}&mode=signup`)
        next()
      } else {
        // login flow!
        const token = await app.passport.createJWT(req.feathers.payload, { jwt: config.jwt, secret: config.secret })
        redirect(res, `${config.google.successRedirect}?token=${token}`)
        next()
      }
    } catch (err) {
      next(err)
    }
  }
}
```

I haven't figured out yet how to do this in 4.0.0, but will be looking into it soon.

KidkArolis:
In response to my own comment above, it's possible to configure scope and prompt for Google like this:

```
authentication:
  oauth:
      google:
        key: abc
        secret: def
        scope: openid
        custom_params:
          prompt: select_account
```

api/authentication/service.md
daffl: I think it should still be `jwtOptions` since it is specifically for creating a JWT (Auth0 and most other libraries are calling it the same, too). I just thought I'd rename the methods to `*accessToken` because it doesn't have to be a JWT.

daffl:
Definitely will have to update the oAuth docs. Going to merge this PR though to keep things going and will make a follow-up with more documentation. Thanks for collecting your issues in https://github.com/feathersjs/feathers/issues/1324, we'll follow up there.

PR #1292, opened 2019-05-08
Use vuepress tabs plugin and replace gitbook codetabs usages

Mairu:
In the crow docs there are still codeblocks from the old gitbook that are "breaking" the documentation.

I added an existent vuepress plugin to add tab support and used it to replace all occurences of codeblocks.

I did not add any additional styling.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -111,5 +111,5 @@ module.exports = {
       }
     ]
   },
-  plugins: ['@vuepress/last-updated']
-}
+  plugins: ['@vuepress/last-updated', 'tabs']
+};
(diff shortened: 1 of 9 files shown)


daffl:
Awesome! I've been meaning to look into this but never got a chance so this is super helpful. Will definitely get this in, just one question: It already looks better in GitHub than the previous Codetabs (I kept getting PRs for trying to fix the strange Codetabs syntax when people looked at it on GitHub) but do you know if there is a way to customize the tab flags? I was thinking instead of e.g.

```
::: tab "Modular"
```

Something like

```
<!-- ::: tab "Modular" -->
```

Would make it not show up at all when viewing on GitHub.

Mairu:
I can try to add comments and will see if it breaks the vuepress output.

Mairu:
So no if you use comments the tabs (and also the content of the tabs) are not displayed anymore when using vuepress.

daffl:
That makes sense. Maybe something for later. This is great though. Thanks again for doing that.

PR #1305, opened 2019-07-10
Add note to "authenticate via existing token" flow

amaanm:
Added a note to the authenticate via an already obtained oauth access token flow description to remind prospective users that they need to add the given oauth strategy to their `authStrategies` config.


daffl:
Good call. Glad it worked as expected otherwise!

PR #1329, opened 2019-08-28
Fix broken links

w3cj:
Found and fixed several broken links using the `validate-links` script.


daffl:
Yes, that's what I noticed, too. Ok, created a follow-up issue in #1336

PR #1341, opened 2019-09-01
update examples to use destructured variables

barandasdemir:
Changed 'context.data.text' to 'data.text' on some examples. Since data is already destructured before, 'context.' is not needed.

edit: thanks to @w3cj, actual suggestion came from him while he was streaming 'going through the getting started guide'


daffl:
Thank you!

PR #1347, opened 2019-09-04
Update oauth.md

edwardsmarkf:
i believe that Crow requires /connect/ in the URL now.   And all that i have tried use "https://" as the only option.


daffl:
Both technically work but it is definitely recommended to rely on the oAuth URLs [already documented in the API](https://docs.feathersjs.com/api/authentication/oauth.html#oauth-urls).

PR #1350, opened 2019-09-04
fix missing async in front of context

devashishsethia:
Second example under asynchronous flow is missing async. Added that.


daffl:
Thank you!

PR #1360, opened 2019-09-17
Update testing.md

MalgAmoe:
NODE_ENV is not test when running test in typescript. It leads to db errors as its using the normal folder.


daffl:
Thank you!

PR #1365, opened 2019-09-24
Update facebook.md

edwardsmarkf:
i believe that name: `profile.name` is required to get the name.


daffl:
Thank you @edwardsmarkf. I don't believe we need the name to get the chat example to work though.

edwardsmarkf:
true - i was more considering the facebook (and google) recipe as a
standalone reference, not necessarily specifically for the chat example.

Thank you,

Mark Edwards

daffl:
Fair, makes sense.

PR #1377, opened 2019-10-07
Clarify info about real-time events

jntesteves:
This part of the documentation was a little misleading. You DON'T need a websocket setup to be able to receive real-time events from the server. The real-time transports via socketio or primus can fallback to http polling in the cases when a websocket connection can not be established. So you can use a realtime transport even if you have some proxy or some other kind network restriction that won't allow a websocket connection.


daffl:
Makes sense. Thanks!

PR #1387, opened 2019-10-17
Use quotes to indicate string values

jordanbtucker:
A few default values for the oauth config were backticked but not quoted, which made it confusing as to whether they were strings or references to other properties.

I also updated the default value for `callback` since it gets its value from `defaults.path`.


daffl:
Thank you!

PR #1393, opened 2019-10-23
Changed ts-node Command in Getting Started

jfavrod:
Used the `ts-node` command that enable me to start the API when using Feathers `4.3.7`.


daffl:
Thank you for the pull request @jfavrod! As per your question earlier I already updated the description in https://github.com/feathersjs/docs/commit/0f722433a19176f68877f18d7fad9ad5b5a72983 to reference `app.ts`. I don't believe the guides mention `index.ts` anywhere else?

PR #1400, opened 2019-11-21
Clarify the note on calling `app.reAuthenticate()`

amxmln:
As discussed [here](https://github.com/feathersjs/docs/issues/1397), this pull request attempts to clarify that it’s safe to call `app.reAuthenticate()` if necessary, but it shouldn’t be called for _every_ request and only very rarely needs to be called more than  once.

Closes https://github.com/feathersjs/docs/issues/1397

### DIFF
--- a/api/authentication/client.md
+++ b/api/authentication/client.md
@@ -51,7 +51,7 @@ app.reAuthenticate().then(() => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


daffl:
Perfect, thank you!

PR #1405, opened 2019-11-30
join(channel) should be join(connection)

devashishsethia:
This is also in the generated channels.js file.


daffl:
Thank you. I'm wondering if this is wrong in the generated file as well.

devashishsethia:
Yes, I mentioned in the pull request.

Regards,
Devashish Sethia

PR #1406, opened 2019-12-03
Fix: (52) Empty reply from server

jeffreyrubi:
If the host is not set to 0.0.0.0, the default "localhost" will rendered into the captioned error when running inside the container.


daffl:
Thank you @jeffreyrubi. I believe with https://github.com/feathersjs/generator-feathers/pull/520 published this may no longer be necessary.

PR #1408, opened 2019-12-16
Fix error in JWT doc

abonventre:
Fixed an issue with the wrong service name in the example for JWT strategy implementation.


daffl:
Thank you!

PR #1439, opened 2020-03-04
Remove NeDB reference in code snippets for file uploads

JorgenVatle:
The Feathers cookbook for file uploads includes a code snippet where NeDB is imported, alongside an explanation for doing so. However, the dependency is never used and may cause confusion to users. It threw me for a bit of a loop. 😅

If this snippet does indeed belong here - perhaps we could refactor to make it more clear where it is necessary? I'd be more than happy to submit an updated PR if this is the case. 👍


daffl:
Thank you!

PR #1451, opened 2020-03-18
fix(docs): SKIP, change link of hooks-common library docs

J3m5:
The link to the hooks common library documentation in the SKIP section point to the old domain.


daffl:
Thank you!

J3m5:
You're welcome, I haven't checked for any other old links, but there might be...

PR #1453, opened 2020-03-23
Corrected typo for windows users

PierreBader:
Relates to https://github.com/feathersjs/docs/issues/1452 
Having a space before the & adds that space at the end of the variable. The NODE_ENV is then set to "test " which is unexpected.


daffl:
Perfect, thank you very much and good catch!

PR #1456, opened 2020-04-06
Fixing param name at FacebookStrategy TS

linzera:
### What this PR do?

 fixes the param name at `FacebookStrategy` `get profile` function:

When I was reading the docs, just copied the snippet `FacebookStrategy` from the TS docs, suddenly I found this inconsistency this PR

### The problem:

- Duplicated identifier (`data`) 
- Also `authResult` does not exists in this scope

authentication ts — scrimmage-api 2020-04-06 01-11-13

### On docs:

Facebook | FeathersJS 2020-04-06 01-15-09


daffl:
Yup, you are absolutely correct. Thank you!

PR #1457, opened 2020-04-08
Add some explanation for the core concepts around channels

optilude:
As per Slack discussion

### DIFF
--- a/api/channels.md
+++ b/api/channels.md
@@ -18,6 +18,18 @@ Some examples where channels are used:
 - Only admins should be notified when new users are created
 - When a user is created, modified or removed, non-admins should only receive a "safe" version of the user object (e.g. only `email`, `id` and `avatar`)
 
+## Concepts
+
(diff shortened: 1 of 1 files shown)


api/channels.md
```diff
@@ -18,6 +18,18 @@ Some examples where channels are used:
 - Only admins should be notified when new users are created
 - When a user is created, modified or removed, non-admins should only receive a "safe" version of the user object (e.g. only `email`, `id` and `avatar`)
 
+## Concepts
+
+When using channels, the server pushes events (such as "created", "removed" etc. for a particular service) down to its clients via *channels*. The client doesn’t listen to these channels directly, but rather subscribes to specific events (e.g. on services) that it is interested in. Those events will only fire if the server pushes relevant data down a channel that the client is connected to.
```
KidkArolis, line 23:
```suggestion
When using channels, the server pushes events (such as "created", "removed" etc. for a particular service) down to its clients via *channels*. The client doesn’t listen to individual channels directly, but rather subscribes to specific events on services that it is interested in. Those events will only fire on the client if the server pushes data to one or more channels that the client has been added to.
```

optilude: Agree with these changes but ... how do I actually accept them into the PR?

daffl:
This definitely makes sense, thank you for writing that up!

PR #1465, opened 2020-04-16
fix(code): remove dollar prompt from command code blocks

soullivaneuh:
This make the copy/paste feature easy to use.

Indeed, if you currently use the copy/paste button, the prompt is also copied, leading to this kind of error:

```
❯ $ npm install @feathersjs/feathers @feathersjs/socketio-client socket.io-client

zsh: command not found: $
```

It not very convenient having to remove this prompt on each command copy.


daffl:
Definitely makes sense especially since it has a copy button on the code sample. Thanks!

PR #1467, opened 2020-04-16
fix: wrong imports for socketio

### LINKED ISSUE #1466: React native io ping interval and timeout looks obsolete
### Comment/Problem

Ref: https://docs.feathersjs.com/api/client.html#react-native

On the first code block, you suggest to configure the client calling `socketio` with the socket object as argument:

```typescript
client.configure(socketio(socket));
```

This works. On the second block, you suggest to adjust the `pingInterval` and `pingTimeout` option to prevent android issues:

```typescript
const app = feathers();
const socketio = require('feathers-socketio');

app.configure(socketio({
  pingInterval: 10000,
  pingTimeout: 50000
}));
```

But:

- The require package is not the same and does not looks to exist anymore (`feathers-socketio`)
- After some search, it looks like the options have to be configured on the `io` socket init directly. But those option are missing. Instead we have `timeout`, `reconnectionDelay` and `reconnectionDelayMax`.

So I assume this should be something like this:

```typescript
const socket = io('https://staging.api.kidways.fr', {
  transports: ['websocket'],
  timeout: 20000,
  reconnectionDelay: 5000,
  reconnectionDelayMax:20000,
});
```

Source: https://github.com/socketio/socket.io-client/blob/8372591652260f88987c467b94573a144ee4db63/docs/API.md#manager

Am I right? And what the exact replacement of `pingInterval` in that case?

I would like to contribute, but I want to be sure of what I am doing before. :-)

soullivaneuh:
fixes #1466

### DIFF
--- a/api/client.md
+++ b/api/client.md
@@ -81,7 +81,7 @@ messageService.create({
 });
 ```
 
-Since React Native for Android doesn't handle timeouts exceeding one minute, consider setting lower values for `pingInterval` and `pingTimeout` of `feathers-socketio` **on your server**. This which will stop warnings related to this [issue](https://github.com/facebook/react-native/issues/12981). For example:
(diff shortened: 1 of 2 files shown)


daffl:
Thank you!

PR #1476, opened 2020-05-04
Instruct tutorial-following-person to create a new file public/app.js

jcjveraa:
I got 'hung up' for a while thinking I needed to modify the server-side app.js while we need to create a new file for the client side.


daffl:
That makes sense. I'm wondering if we should name it `public/client.js` instead since it appears there has been some confusion before.

PR #1491, opened 2020-07-09
Explain that user service is not yet ready

MarcelWaldvogel:
This is to avoid giving the impression that the user service can already be experimented with now.

### DIFF
--- a/guides/basics/services.md
+++ b/guides/basics/services.md
@@ -281,6 +281,8 @@ Now we can sign up users with email and password and it will automatically set a
 
 > __Note:__ We are keeping `githubId` from the original data so that we can add a "Login with GitHub" button in the [authentication](./authentication.md) chapter.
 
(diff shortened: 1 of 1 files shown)


daffl:
Thank you @MarcelWaldvogel for the pull request. I'm not sure I am following though. At the end of this chapter you should be able to create a new user without problems.

MarcelWaldvogel:
Other tutorials for other web frameworks I have gone through in the past try to give you a minimal working UI at the end of each chapter, whenever possible. Here, I was expecting something similar, but only the REST API is available.

Ah, I see, my wording is raising (or, rather, lowering) more expectations than necessary.

Would rephrashing to:

> Please be patient and don't expect to be able to create users **through a UI** yet; even though we have a **REST** service, it is not yet ~~ready and~~ integrated.

be better?

daffl:
I'm not sure. The [Quick start guide](https://docs.feathersjs.com/guides/basics/starting.html#installing-feathers) already shows the whole flow from creating the server and a simple UI for integrating it.

PR #1492, opened 2020-07-10
Update testing.md

anonimusprogramus:
Note for adding .nyc_output/ folder to .gitignore


daffl:
Makes sense. We might even want to add it to the `.gitignore` created by the generator.

PR #1493, opened 2020-07-14
Updated configuration docs for nested properties

bennol:
Added important detail that when using @feathers/configuration,  app.get() can't directly access nested properties. 
https://github.com/feathersjs-ecosystem/configuration/issues/38

I stumbled upon this unexpected behavior and spent some time debugging the issue. Having worked with node-config for some time I just expected nested properties to work just as well with `app.get()`/`app.set()` as with `config.get()`/`config.set()`.


daffl:
Thank you!

PR #1496, opened 2020-07-25
fix: use consistent reference to OAuth

calendee:
There are several instances of the docs referencing both `oAuth` and `OAuth`.

This PR changes all references to `OAuth`: https://en.wikipedia.org/wiki/OAuth


daffl:
Thank you!

PR #1498, opened 2020-07-26
docs: add hook-less service method description

bwgjoseph:
Currently, the only place learn about hook-less service method is at https://docs.feathersjs.com/guides/migrating.html#hook-less-service-methods and it isn't obvious when looking through the docs..

I have added it to `api/databases/common` at the top after `constructor` but not sure if it should be at the bottom after `adapter.remove(id, params)`. What do you think?

This was brought up by Mike from slack...

### DIFF
--- a/api/databases/common.md
+++ b/api/databases/common.md
@@ -179,6 +179,18 @@ This section describes specifics on how the [service methods](../services.md) ar
 
 Initializes a new service. Should call `super(options)` when overwritten.
 
+### hook-less
+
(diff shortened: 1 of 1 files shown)


mikestefanello:
The hook-less service methods were useful for me because I was attempting to load the requested object from with hooks, and it would result in an infinite loop:

```
const obj = await context.service.get(context.id);
```

Using `_get` avoids that.

daffl:
Great call, thank you!

PR #1499, opened 2020-07-29
Bumped up feathers & socketio client version

shihabmridha:
Bumped up FeathersJS and SocketIO client to the latest version in "Load from CDN with <script>" section.

### Background:
I was experimenting with custom events. I just copied the snippet from the [documentation](https://docs.feathersjs.com/api/client.html#load-from-cdn-with-script). But, when I tried to call `app.reAuthenticate()` it didn't work. Then I noticed that I am importing old version (`3.0.0`) of Feathers client (_the `^` in the URL didn't import the latest one_). So, I changed the version to the latest (`4.5.7`) one and it worked as expected. I've updated the version of socketio as well (`1.7.3` to `2.3.0`).


daffl:
Thank you!

PR #1505, opened 2020-10-13
Update socketio.md

keliq:
fix typo


daffl:
I guess it's valid HTML5 but you're right, it's the same everywhere else so definitely makes sense for consistency.

PR #1506, opened 2020-11-07
Cookbook for Firebase Authentication

GioLogist:
Excited to officially have this in the docs, since I always find myself re-writing different versions for new projects 🤙🏻

In order to remain unopinionated on a custom design, I decided to leverage Firebase's official [Firebase UI](https://firebase.google.com/docs/auth/web/firebaseui). However, it would be fairly easy for anyone to create an entirely custom design on the client side. While still following these docs.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -100,6 +100,7 @@ module.exports = {
           'authentication/auth0.md',
           'authentication/facebook.md',
           'authentication/google.md',
+          'authentication/firebase.md',
           'authentication/_discord.md',
           'authentication/stateless.md',
           'authentication/revoke-jwt.md'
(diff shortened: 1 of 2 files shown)


GioLogist:
Let me know if there's anything I overlooked! 

I'm happy to get a demo up and running ASAP as well. I wasn't sure if it was 100% necessary, but I can certainly see the value, since this is a bit involved

daffl:
Awesome, thank you for writing this up! I'll merge it and we can see if a demo will be necessary or not. Usually I found not if the document contains all the information necessary (often it can even be in the way because there is then two places where things need to be updated when anything changes).

GioLogist:
@daffl Great point! On stand-by if needed 🤙🏻

PR #1512, opened 2021-01-11
Correct syntax of $in example in querying.md

devashishsethia:
$in expects an array. So, this breaks if there is a single item. e.g. GET GET /messages?roomId[$in]=2 will fail.
Please see here:
https://feathersjs.slack.com/archives/C0HJE6A65/p1610228957226400


daffl:
Thank you!

PR #1513, opened 2021-01-19
Update link to bCryptJs & update `hashSize` definition

dyllan-to-you:
- Added link to bCryptJS npm repository
- The previous definition of `hashSize` stated that it was the "BCrypt Hash Size". This is both inaccurate and misleading; the parameter appears to represent the length of the salt used to hash the passwords. The variable name is also misleading, but that requires a bigger fix & potentially breaking change


daffl:
Well, a link definitely makes sense. Thanks!

PR #1518, opened 2021-02-19
Tell Typescript users to modify tsconfig

SephReed:
In order for a typescript project to work, tsconfig needs to be modified.  Added a section telling users to do so.


daffl:
Thank you for the pull request. This has now been addressed in v5.0.0-beta.1 which has many TypeScript improvements and does no longer require synthetic imports. See the migration guide for [testing the prerelease](https://dove.docs.feathersjs.com/guides/migrating.html#testing-the-prerelease).

PR #1519, opened 2021-02-28
Updating docs to indicate node pm2 also uses `username` environment variable if not escaped

meditatingdragon:
Updating documentation informing users to escape userField: `\\username` if running Feathers with PM2, as it also just pulls the local environment variable username instead of the usernameField.


daffl:
Thank you! Makes sense to mention in here for now - and another reason why this kind of environment variable substitution will be removed in the upcoming version.

PR #1520, opened 2021-03-17
Options object passed into wrong method

philipdalen:
The options object was passed in to the wrong method the rest method does not except more than one parameter.


daffl:
Thank you!

PR #1523, opened 2021-03-29
Fix typo in params.headers definition

jordanbtucker:
> The HTTP headers connected to this service call if available. This is either the headers of the REST call of the headers passed when initializing a websocket connection.

I believe the sentence should say "REST call *or* the headers".


daffl:
It does indeed, thank you!

PR #1524, opened 2021-04-01
add possible missing step

edwardsmarkf:
add possible missing step

### DIFF
--- a/guides/migrating.md
+++ b/guides/migrating.md
@@ -7,6 +7,7 @@ This guide explains the new features and changes necessary to migrate to the Fea
 You can run the following to test the latest Dove pre-release in your application:
 
 ```
+npm install -g npm-check-updates;  ## if not already done
 npx npm-check-updates --upgrade --target newest --filter /@feathersjs/
 npm install
 ```


daffl:
Thank you for the pull request. Did you get an error running just the command? `npx` should normally do the installation automatically.

edwardsmarkf:
i get:
# npm-check-updates;
-bash: npm-check-updates: command not found

Thank you,

Mark Edwards

daffl:
The command is `$ npx npm-check-updates` (note the [npx](https://docs.npmjs.com/cli/v7/commands/npx))

edwardsmarkf:
yes i did that and got the error i showed you before.  😁  but maybe
something might be wrong in my setup....?

imho, i would also terminate every command with a semicolon for
readability sake so that way it's clear if this is a new command or an
extension of the previous command.  but again, imho.

happy April 1st to you!

Thank you,

Mark Edwards

daffl:
What happens if you run something like `npx http-server`?

edwardsmarkf:
#  npx http-server
(node:30132) ExperimentalWarning: The fs.promises API is experimental
Need to install the following packages:
  http-server
Ok to proceed? (y) y
sh: http-server: command not found
npm notice
npm notice New patch version of npm available! 7.7.4 -> 7.7.6
npm notice Changelog: https://github.com/npm/cli/releases/tag/v7.7.6
npm notice Run npm install -g ***@***.*** to update!
npm notice


Thank you,

Mark Edwards

daffl:
It does look like something with your setup - on MacOS and Linux I definitely recommend using [nvm](https://github.com/nvm-sh/nvm) - installed normally with local permissions - with one of the later Node versions.

edwardsmarkf:
for fun i did the following - looks like you are correct:

1) google-vm centos-8

2-6 from  https://gist.github.com/leandrino/e84ca31e69e7803bcf7d91996316cc79


2) dnf -y update ;

3) dnf install -y nodejs npm ;

4) npm install -g nvm ;
               #npm WARN deprecated ***@***.***: This is NOT the correct nvm.
Visit http://nvm.sh and use the curl command to install it.

 5 from https://github.com/nvm-sh/nvm/blob/master/README.md
5) curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh
| bash ;

npm install -g n  ;

6)  npx -v ;
6.14.11

7)  npm -v ; node -v  ;
6.14.11
v10.24.0

8)  npx http-server ;
npx: installed 30 in 5.494s
Starting up http-server, serving ./
Available on:
  http://127.0.0.1:8080
  http://10.128.0.57:8080
Hit CTRL-C to stop the server

Thank you,

Mark Edwards

daffl:
Yup, that's what `npx http-server` should show. The command from the docs would then also work. So it's likely some misconfiguration on your system. Common ones are wrong permissions or a wrong `PATH`

edwardsmarkf:
ah gotcha - so move off of centos-7 and on to centos-8, or better yet,
rocky-linux (even if its been delayed a month).😊

Thank you,

Mark Edwards

PR #1527, opened 2021-04-12
Update apiKey.md

ajmeyghani:
Reading the token from the query parameters can cause confusion, especially when it comes to `get` and `find` calls if you are not careful. You have to set up a hook to check for a specific query parameter and remove it before it's called by Feathers. It's just better to read if from the request headers, but if you have a specific use cause to read the token from query params, then you have to filter it out properly before the call is made by Feathers.


daffl:
Thank you!

PR #1529, opened 2021-04-30
Initial API documentation pages for KoaJS adapter

daffl:
For https://github.com/feathersjs/feathers/pull/2315

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -44,7 +44,6 @@ module.exports = {
       }],
       '/api/': [{
         title: 'Core',
-        collapsable: false,
         children: [
           'application.md',
           'services.md',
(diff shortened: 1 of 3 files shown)


bertho-zero:
I see that the Koa transporter does not allow to receive an existing app as a second parameter.

And that the express one is not documented.

https://github.com/feathersjs/feathers/blob/a43e7da22b6b981a96d1321736ea9a0cb924fb4f/packages/express/src/index.ts#L24

https://github.com/feathersjs/feathers/blob/a43e7da22b6b981a96d1321736ea9a0cb924fb4f/packages/koa/src/index.ts#L17

daffl:
I believe they are both documented at https://docs.feathersjs.com/api/express.html#express-2 and https://dove.docs.feathersjs.com/api/koa.html#koa-2 it should work pretty much the same (although Koa needed the Object assignment the other way around for inheritance reasons).

bertho-zero:
I don't see any mention of the second argument to use an existing express or koa app, the links you send me show the signature with no arguments.

daffl:
Oooh... sorry I forgot about that, that is a good point.

daffl:
Added to the docs via https://github.com/feathersjs/docs/commit/76a8af41cf8ed6a6a4514470abcbc5314fb9f527

PR #1531, opened 2021-06-02
Fix standardJS errors (no-undef)

e-dot:
In order to avoid errors "no-undef" from standardJS validator, I had to add global variables in client.js, like this:

/* global io, feathers, moment */


daffl:
That makes sense, thank you!

PR #1532, opened 2021-06-02
Fix errors on new installation : Cannot find module ...

e-dot:
Fix errors on a brand new installation:
* Cannot find module '@feathersjs/authentication'
* Cannot find module '@feathersjs/authentication-local'
* Cannot find module '@feathersjs/authentication-oauth'


daffl:
Hi @e-dot. Thank you for the pull request! These modules will be installed for you automatically when you generate a new application and should not be installed globally. If you received an error, try

- Removing `package-lock.json` and the `node_modules` folder, then run `npm install` again
- Check if you got an error while generating the application or running `npm install`

PR #1535, opened 2021-07-20
Update jwt.md

indatawetrust:
i think the correct value should be jwtOptions instead of jwt

https://github.com/feathersjs/feathers/blob/687e3c7c36904221b2707d0220c0893e3cb1faa9/packages/authentication/src/options.ts#L3


daffl:
Thank you!

PR #1540, opened 2021-08-29
Add forgot password

DaddyWarbucks:
Let me know if you have any questions/concerns. I am happy to update anything.

### DIFF
--- a/cookbook/authentication/forgot-password.md
+++ b/cookbook/authentication/forgot-password.md
@@ -78,10 +78,7 @@ module.exports.Service = Service;
 // services/reset-password.class.js
 
 const jwt = require("jsonwebtoken");
-const util = require("util");
 const { BadRequest } = require("straightline-utils/errors");
-
-const verifyToken = util.promiseify(jwt.sign);
 
 class Service {
   setup(app) {
@@ -141,7 +138,7 @@ class Service {
     const resetSecret = this.app.get("resetSecret");
 
     try {
-      await verifyJWT(passwordResetToken, resetSecret);
+      jwt.verify(passwordResetToken, resetSecret);
     } catch (error) {
       throw new BadRequest("Password reset token is invalid.");
     }
--- a/cookbook/authentication/forgot-password.md
+++ b/cookbook/authentication/forgot-password.md
@@ -1,0 +1,167 @@
+# Forgot Password
+
+This cookbook is a basic example that shows how to create a `forgot-password` and `reset-password` service. Because resetting passwords requires sending emails (and/or texts), additional model properties and config, etc, it is not covered by the official feathers-authentication library. Instead, you are encouraged to use the following example and change it as needed.
+
+Prerequisites
+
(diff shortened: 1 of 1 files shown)


cookbook/authentication/forgot-password.md
```diff
@@ -0,0 +1,139 @@
...
+    //   query: { passwordResetToken }
+    // });
+
+    if (!user) {
+      return { valid: false };
+    }
+
+    const resetSecret = this.app.get("resetSecret");
+
+    return verifyToken(passwordResetToken, resetSecret)
```
ajmeyghani, line 84: @DaddyWarbucks  quick question, is this a wrapper for `jwt.verify`? thanks

DaddyWarbucks: Yep. Looks a typo from my copy/paste job. 👍

DaddyWarbucks:
For a more robust/safe solution see: https://blog.logrocket.com/implementing-a-secure-password-reset-in-node-js/
Perhaps this PR should be updated to look more like this article.

cookbook/authentication/forgot-password.md
```diff
@@ -0,0 +1,139 @@
...
+  async get(passwordResetToken, params) {
+    const [user] = await this.usersService.find({
+      query: {
+        passwordResetToken,
+        $limit: 1,
+      },
+      paginate: false,
+    });
+
+    // Pro Tip - Use findOne(). See findOne cookbook.
```
bartduisters, line 73: If it is a pro tip, why not use `findOne()` instead of `find()` in the implementation?

cookbook/authentication/forgot-password.md
```diff
@@ -0,0 +1,139 @@
...
+    this.app = app;
+    this.usersService = app.service("api/users");
+    this.emailService = app.service("mailer");
+  }
+
+  async create(data, params) {
+    const { email } = data;
+
+    const resetSecret = this.app.get("resetSecret");
+    // const resetSecret = this.app.get('authentication').secret;
```
bartduisters, line 28: I would add an example json used in the config/ directory and only keep the relevant one, delete the comments.

cookbook/authentication/forgot-password.md
DaddyWarbucks: Because `findOne` is not a standard method. You would need to implement it yourself. I don't think there is actually even a "findOne cookbook". Keep in mind this PR was a result of a thread in the Slack channel and was added here to capture a potential solution and kinda document that conversation. I basically just copy/pasted one of my implementations as an example and threw in some comments. There are a number of reasons it is not merged. Like I said in my previous comment, this PR should probably be updated to look something like  https://blog.logrocket.com/implementing-a-secure-password-reset-in-node-js/.

We also discussed a generic "token service" (similar to the one in that article) that a developer could use for all kinds of things.

bartduisters: Ok, clear!

daffl:
Apologies for not getting to this earlier. We can totally get this in, only thing left is to add this file to the navigation in https://github.com/feathersjs/docs/blob/crow/.vuepress/config.js#L105

PR #1545, opened 2021-09-23
fix helmet compatibility and remove innerHTML use in starter code

FossPrime:
Starter code has the following issues:
- inline attribute script, which the default feathers(helmet()) call disallows
- innerHTML which will break any event listeners in #493
    - injection hazard

# Screenshot of this exact code actually working...

Note:
- The server is running dove. The client is unchanged 4.3
- Messages are still not sorted, leaving that as exercise for the reader

Screenshot from 2021-09-22 21-07-11

### DIFF
--- a/guides/basics/starting.md
+++ b/guides/basics/starting.md
@@ -459,7 +459,7 @@ Now we can look at one of the really cool features of Feathers. It works the sam
       const messages = await app.service('messages').find();
 
       // Add existing messages to the list
-      messages.data.forEach(addMessage);
+      messages.forEach(addMessage);
 
       // Add any newly created message to the list in real-time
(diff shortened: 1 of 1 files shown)


daffl:
Thank you for the pull request. I don't think the `messages.data` change is correct though. The service defined in the [An API server](https://docs.feathersjs.com/guides/basics/starting.html#an-api-server) section returns `this.messages` in `find` which is an array.

FossPrime:
> Thank you for the pull request. I don't think the `messages.data` change is correct though...

I thought feathers-memory was as simple as it got. Good to know.

daffl:
It is but this one is a custom service 😄

PR #1555, opened 2022-01-22
the link to middleware of sockeio was broken

UtmostCreator:
(404 - not found) is thrown when clikcing on 
`Socket.io middleware` ... (opens new window)can modify the feathers property on the socket which will then be used as the service call params:


daffl:
Thank you!

PR #1557, opened 2022-01-31
Fix errors in the Feathers guide

AtomicTroop:
The TypeScript version of the guide has some missing "any" types in the tests added during the final section of the guide.
This pull request fixes those omissions, matching the current state of [feathers-chat-ts](https://github.com/feathersjs/feathers-chat-ts).


AtomicTroop:
I notice the mocha script in the "Writing tests" section of the guide uses ts-mocha while the feathers-chat-ts repo uses a more complex mocha command that matches and builds on what the current feathers project generator produces in package.json by default. Not sure if that should also be fixed?

PR #1563, opened 2022-03-15
Fix browser `fetch` configuration example

strarsis:
The browser `fetch` configuration example (FeathersJS REST client) doesn't bind `window` to `fetch`,
the `fetch` used by the FeathersJS REST client won't have the right `this` context of `window`, 
which causes a JavaScript error:
````
Uncaught (in promise) Error: Failed to execute 'fetch' on 'Window': Illegal invocation
````

Related issue: https://github.com/josx/ra-data-feathers/issues/177

### DIFF
--- a/api/client/rest.md
+++ b/api/client/rest.md
@@ -70,6 +70,8 @@ const messages = app.service('messages');
 
 <!-- -->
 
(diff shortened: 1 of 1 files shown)


daffl:
Interesting, I'm curious when this changed. Thanks for putting it up!

PR #1566, opened 2022-04-12
Fix client imports in dove docs

FossPrime:
Imports were outdated.

https://stackblitz.com/edit/vitejs-vite-gx8b9j?file=public%2Findex.html&terminal=dev


daffl:
This is fixed now in https://feathersjs.com/guides/basics/starting.html

PR #1571, opened 2022-05-11
fix: remove app.listen(..).on broken function

YvesMolina:
When I tried to run the code for the app.js in the "An API Server" part of the tutorial, I got this error : 
TypeError: app.listen(...).on is not a function

I am using node v16.13.2 and npm 8.7.0.

I replaced the function causing the error with a working syntax and the app launched correctly.


daffl:
Thank you for the pull request! This probably happened because the v5 prerelease was accidentally published as latest for a short time. The original example should work again (and will be updated accordingly for the official release).

PR #1572, opened 2022-05-15
update usage example in dove for koa

gobengo:
Motivation
* I copy-pasted the old way with `app.use(rest())` and it led to a runtime http response error like `app.use is not a function`. I think it's because `rest()` is expected to be passed to `app.configure()` and not `app.use()`.


daffl:
That did indeed change. Thank you!

